PR #110, opened 2022-08-05
Update dpl build workflow and switch to multiarch

danquah:
#### What does this PR do?
Updates the github action workflow that builds, tests and releases dplsh.

This update
- Reformats the file
- Fixes a bug where we triggered on the nonexistent "master" branch (should be main)
- Simplified the workflow by using a docker meta action for generating labels and tags
- Switched to buildx to do multiarch builds

The action tests itself, I've temporarily allowed it to always push so the runs made by this PR shuold create a full reelase.

### DIFF
--- a/.github/workflows/dplsh-build-release.yaml
+++ b/.github/workflows/dplsh-build-release.yaml
@@ -6,70 +6,91 @@
 on:
   pull_request:
     paths:
-    - 'tools/dplsh/**'
-    - '.github/workflows/dplsh-build-release.yaml'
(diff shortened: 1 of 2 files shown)


achton (approved):
LGTM

There is [a warning from `git`](https://github.com/danskernesdigitalebibliotek/dpl-platform/runs/7689114668?check_suite_focus=true#step:22:15) during action runs but it is not new for this change. Does not appear to affect the build either.

PR #118, opened 2022-08-31
Fix cluster:auth task

achton:
#### What does this PR do?

Since we are using `terraform -chdir=` to invoke terraform from an env subdir, we do not also need to ask Task to chdir for us first.

This fixes that behaviour.


danquah (approved):
Ahhh, that makes sense - good catch :)

PR #119, opened 2022-08-31
Bump limit for dev environments

achton:
#### What does this PR do?

This will bump the environment limit for all new projects when we create them, from 5 to 25.

This is probably a more sane default than the current upstream one.


danquah (approved):
👍

PR #143, opened 2022-11-07
Improve Kubernetes access docs

kasperg:
#### What does this PR do?

Improve documenation for accessing Kubernetes.

There are a number of steps/rrerequisites which only needs to be setup once and are thus easy to forget. Let's document them!

Good question. I bet you do not want to reset your entire development environment for this so I guess a sanity check is OK.

### DIFF
--- a/documentation/runbooks/access-kubernetes.md
+++ b/documentation/runbooks/access-kubernetes.md
@@ -8,6 +8,8 @@ When you need to gain `kubectl` access to a platform-environments Kubernetes clu
 
 * An authenticated `az` cli. The logged in user must have permissions to list
   cluster credentials.
+* `docker` cli which is [authenticated against the GitHub Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry).
+  The access token used must have the `read:packages` scope.
 
 ## Procedure
 
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -34,7 +34,13 @@ tasks:
 
     # Internal task for initializing Terraform.
     _infra:terraform:init:
+      deps:
+        - _req_env
       dir: "{{.dir_infra}}"
+      sources:
+        - "*.tf"
+      generates:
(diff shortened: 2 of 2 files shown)


documentation/runbooks/access-kubernetes.md
```diff
@@ -8,6 +8,8 @@ When you need to gain `kubectl` access to a platform-environments Kubernetes clu
 
 * An authenticated `az` cli. The logged in user must have permissions to list
   cluster credentials.
+* `docker` cli which is [authenticated against the GitHub Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry).
```
danquah, line 11: I'm not sure: did we still need this or did the need go away with the update to `.lagoon.yml` to no longer reference the regsitry?

kasperg: This is still needed to get [the `dplsh` image](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/main/tools/dplsh/dplsh.sh#L22).

infrastructure/Taskfile.yml
```diff
@@ -93,17 +99,25 @@ tasks:
...
+        # We normally handle variables via dynamic task variables, but as terraform.
+        # Variables are interpolated before dependencies are resolved, so
+        # in cases where Terraform is not initialized, this task will break.
+        # A lot of other tasks are in the exact same situations.
+        # So, to mitigate this, we instead export the variables we need as
+        # commands.
+        # We sneakily rely on the fact that most procedures starts off with a
+        # cluster:auth, which means most people will have their terraform workspace
+        # initialized before the run the next task which can then extract variables
+        # the "normal" way.
```
kasperg, line 115: :+1: nice.

documentation/runbooks/access-kubernetes.md
danquah: Ahh, got it. We should probably also add it to https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/main/documentation/runbooks/using-dplsh.md then

PR #147, opened 2022-11-30
Upgrade Lagoon and AKS

danquah:
#### What does this PR do?
I've upgraded the cluster to kubernetes veersion 1.24.6 and lagoon to 2.11.0

The upgrade revealed that we had a need for a more nuanced handling of versions of our node-pools so I updated the Terraform setup to allow us to control the versions of the individual nodepools.

I also noticed we where using a silly volume size for harbors pvcs and updated those.

This PR is made after the fact, so the review is mostly about reading through the updated documentation and the changes to Terraform.

DDFDRIFT-1
DDFDRIFT-14

### DIFF
--- a/documentation/runbooks/upgrading-aks.md
+++ b/documentation/runbooks/upgrading-aks.md
@@ -62,11 +62,34 @@ task infra:provision
 
 ### Upgrade the cluster
 
-1. Then update the `kubernetes_version` reference in `infrastructure/environments/<environment>/infrastructure/main.tf`
+Upgrade the control-plane:
+
+1. Update the `control_plane_version` reference in `infrastructure/environments/<environment>/infrastructure/main.tf`
   and run `task infra:provision` to apply. You can skip patch-versions, but you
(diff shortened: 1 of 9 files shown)


documentation/runbooks/upgrading-aks.md
```diff
@@ -62,12 +62,35 @@ task infra:provision
...
+2. Monitor the upgrade as it progresses. Expect the provisioning of and workload
+   scheduling to a single node to take about 5-10 minutes. In particular be aware
+   that the admin node-pool where harbor runs has a tendency to take a long time
+   as the harbor pvcs are slow to migrate to the new node.
+
+Monitor via eg.
 
 ```shell
-watch -n 5 kubectl get nodes
+watch -n 5 get nodes
```
spaceo, line 95: Why is kubectl removed here?

spaceo (commented):
LGTM. Have one question though

documentation/runbooks/upgrading-aks.md
danquah: hah, good catch

PR #163, opened 2023-02-28
Fix the ownership of the files the entrypoint rsyncs into the container.

danquah:
#### What does this PR do?
Fix the ownership of the files the entrypoint rsyncs.

This required a change where the entrypoint now runs as root, but drops privilegeds when it is done with its work.

Do a local build of dplsh and see if you can start it.


danquah:
Co-developed with @oliver-reload - merging.

PR #164, opened 2023-03-06
Updated procedure for support workload update

danquah:
- Fix a bug where DIFF would not work if set as an argument to Task
- Initial work on a runbook for upgrading support workload


danquah:
Self-reviewing - the changes are minimal, and I'm going to follow up with more work on the runbook shortly where a reviewer will get a chance to critic the work.

PR #175, opened 2023-04-05
Implement automation for extracting the current and latest version of support workloads

danquah:
#### What does this PR do?
We want to keep track of how far behind we are on keeping things up to date. This can be a somewhat manual process
so to lessen the toil this commit implements an automated process for going through the things we have currently
deployed and getting the current and latest version of the projects.

We're currenly only doing this for the support workloads, and not for Lagoon and AKS.

It should be possible to simply follow the newly created runbook

no

DDFDRIFT-26


danquah:
(just ignore the branch-name - i got two changes mixed up 😆 )

oliver-reload:
I noticed that `task ops:get-versions` returns minio version twice, but ran out of time while investigating;
```
- bitnami/minio:
    current: 11.2.15
    latest: 12.2.4
- bitnami/minio:
    current: 11.2.15
    latest: 12.2.4
```
I edited conditional formatting in the versions sheet as Cell C6 was affected by changes to D6.

Other than that; LGTM :+1:

oliver-reload (approved):
https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/175#issuecomment-1505194869

danquah:
Hah, you're right about mino - it's because of [this](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/edcada2332b778f553928f472ec6d2ddd4d45007/infrastructure/Taskfile.yml#L196) - we have two installations.

It should be ok for now, I've done a quick verify with the latest changes from main and everything still seems to work, I'll merge.

PR #179, opened 2023-04-13
Update grafana, cert-manager. Improve docs and tools

danquah:
#### What does this PR do?
I'm working on streamlining our update procedure, and this is one of many steps towards that goal.

I've updated cert-manager and grafana, and documented the process. During my work I also implemented
a simple update-feature in dplsh, made is possible to run terraform init, and added a tool to dplsh
that allows us to verify cert-manager.

Mostly review of the docs, but you can also

- Check out the branch and verify that running `dplsh.sh --update` pulles the latest version of the image
- See that `task support:verify:cert-manager` works inside a shell

Mostly, just read through it and consider if it is in a state where you would be ok with taking the reigns.

### DIFF
--- a/documentation/runbooks/upgrading-support-workloads.md
+++ b/documentation/runbooks/upgrading-support-workloads.md
@@ -18,13 +18,17 @@ TODO
 
 TODO
 
-## Procedure
+## General Procedure
 
 * Identify the version you want to bump in the `environment/configuration` directory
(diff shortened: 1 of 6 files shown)


oliver-reload (approved):
LGTM. 
I upgraded successfully and support:verify:cert-manager works.

PR #224, opened 2024-02-07
Refresh 2024/2

kasperg:
#### What does this PR do?

This is result of work done by @spaceo and I when setting up new pilot libraries.

I guess not. The changes are the result of actual work needed (or not) when setting up the pilot libraries.

I suggest going through the commits one by one.


achton (approved):
LGTM 👍🏻

PR #225, opened 2024-02-07
Update lagoon images. The previous version is 2 years old.

spaceo:
The old  [lagoon images ](https://github.com/uselagoon/lagoon-images/tree/main)version breaks with the php 8.1 images
The lagoon images version tag we use in dpl-cms is just latest:
https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/develop/lagoon/cli.dockerfile#L1


kasperg (approved):
:+1: Consider creating an link in code comments between this version and what is defined in [`docker-compose.yml`](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/develop/docker-compose.yml).

PR #241, opened 2024-03-07
Guide adjustments

hypesystem:
Draft because this is WIP as we run through runbooks and try to use them.

- Lagoon ssh keys as a prerequisite now also list where to create those keys
- When creating a site, a git-url must be provided - it is now clarified that the repo it points to does not need to exist before running the tasks in the runbook.

Read and verify the changes improve clarity, while staying as concise as possible.

N/A


spaceo (approved):
LGTM 👍

PR #242, opened 2024-03-07
Back up at midday and again in the evening

ITViking:
#### What does this PR do?
DDF wants two daily backups to be made; one around lunch and another in the evening. 
This PR changes the backup scheduling from running in the late evening to running at 12:00 and again at 19:00.

I don't see any good path to test this outside of the production envs. We should however have a look at the backup being made the day this PR is merged in. 

Ensure that the times entered at the hours place is 12 and 19.

https://reload.atlassian.net/browse/DDFDRIFT-35

### DIFF
--- a/infrastructure/dpladm/env-repo-template/standard/.lagoon.yml
+++ b/infrastructure/dpladm/env-repo-template/standard/.lagoon.yml
@@ -82,4 +82,4 @@ backup-retention:
     hourly: 0
 # https://docs.lagoon.sh/lagoon/using-lagoon-the-basics/lagoon-yml#backup-schedule
 backup-schedule:
-  production: M 12,19 * * *
+  production: M H(12-14),H(19-21) * * *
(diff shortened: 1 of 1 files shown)


infrastructure/dpladm/env-repo-template/standard/.lagoon.yml
```diff
@@ -82,4 +82,4 @@ backup-retention:
     hourly: 0
 # https://docs.lagoon.sh/lagoon/using-lagoon-the-basics/lagoon-yml#backup-schedule
 backup-schedule:
-  production: M H(22-2) * * *
+  production: M 12,19 * * *
```
hypesystem, line 85: May be worth considering keeping the jitter (c.f. [what H does in cron syntax](https://stackoverflow.com/a/47302727/1080564)) here?

spaceo (commented):
What about the daily rentention rate?
Should it be doubled now the backup frequency has been doubled?

infrastructure/dpladm/env-repo-template/standard/.lagoon.yml
ITViking: yes, might be a good idea aswell. It does jitter on the minute though

ITViking:
We actually talked about that as well. We should probably double it, to give ourselves enough time :)

ITViking:
@spaceo i made the requested changes

hypesystem (approved):
Looks good to me

Would like @spaceo's final approval too, though, just because this is new stuff for us to be touching :smile:

spaceo (approved):
👍 LGTM

PR #254, opened 2024-03-21
Try this format instead?

spaceo:


hypesystem:
Something that confuses me a bit is that this was the original format:

https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/242/files#diff-271f4a22d7a8ff4430bc7909bda401fed66d96182b7302469652ed8bb8e096e3L85

Which we seem to have moved far away from - it had the `H(x-y)` syntax *and* prefixed "M". Weird that we couldn't make that work with twice-per-day backups?

ITViking:
Did this change work @spaceo? According to Lagoons documentation we're supposed to leave the first Placement as M so Lagoon can decide on the minute to start the backup on.

PR #260, opened 2024-04-05
Revert to backing up once per day

ITViking:
--


hypesystem (approved):
Syntax looks good ! If the backups work let's get it merged

spaceo (approved):
Guess it has been tested ;) LGTM 👍

hypesystem (changes requested):
Hmm, I don't think this actually works! It is taking *a lot* of backups on canary, which is where we were testing it as far as I recall?

https://ui.lagoon.dplplat01.dpl.reload.dk/projects/env-canary/env-canary-main/backups

(Odd as this cron string was approved by Amazee's folks!)

hypesystem (approved):

hypesystem:
fyi @spaceo we are rolling back to backup once per day so it actually *works* - right now backups are every 10 minutes on every site.

A new PR after this will try to reintroduce twice-daily backups

PR #275, opened 2024-04-23
Provision mail services

ITViking:
#### What does this PR do?
This PR adds a mailing service

This PR can be tested by running the following comand in dpl shell: 
`PROJECT_NAME=struer task lagoon:project:ensure:azure-mail-connection-string` and check that the variable has been set. It's possible to view the value of the variable by pressing "show values" 
image

Please test it on fx. the struer project

https://reload.atlassian.net/browse/DDFDRIFT-60

### DIFF
--- a/infrastructure/environments/dplplat01/lagoon/lagoon-remote-values.template.yaml
+++ b/infrastructure/environments/dplplat01/lagoon/lagoon-remote-values.template.yaml
@@ -22,9 +22,6 @@ sshCore:
 dbaas-operator:
   enabled: true
 
-azureEmailService:
-  connectionString: "$ACS_CONNECTION_STRING"
-
   mariadbProviders:
     production:
       environment: "production"
--- a/infrastructure/environments/dplplat01/lagoon/lagoon-remote-values.template.yaml
+++ b/infrastructure/environments/dplplat01/lagoon/lagoon-remote-values.template.yaml
@@ -22,6 +22,9 @@ sshCore:
 dbaas-operator:
   enabled: true
 
+azureEmailService:
+  connectionString: "$ACS_CONNECTION_STRING"
+
   mariadbProviders:
     production:
       environment: "production"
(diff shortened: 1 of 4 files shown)


infrastructure/environments/dplplat01/lagoon/lagoon-remote-values.template.yaml
```diff
@@ -22,6 +22,9 @@ sshCore:
 dbaas-operator:
   enabled: true
 
+azureEmailService:
+  connectionString: "$ACS_CONNECTION_STRING"
```
hypesystem, line 26: This is nested wrong afaict. Should be above `dbaas-operator`. The `mariadbProviders` field *must* be nested in `dbass-operator` and adding thesel ines here breaks that.

hypesystem: Notice also that the lagoon-remote-values.yaml generated from this template is used to helm install a chart.

That means only values known by that chart have any effect. Is `azureEmailService` in the `lagoon/lagoon-remote` chart? (Find out by running `helm show values` on it.)

Latest version of the chart is here and it has no email field: https://github.com/uselagoon/lagoon-charts/blob/main/charts/lagoon-remote/values.yaml

hypesystem: I think we probably need to pass the values in elsewhere, not via lagoon-remote. Lagoon-remote is lagoon's way of running standard drupal stuff. We just need to pass an arbitrary environment value in, preferably as a secret.

We could use lagoon's built in variable support and set it on each project, see e.g. the two `LAGOON_` variables set and passed into dpl-cms here: https://ui.lagoon.dplplat01.dpl.reload.dk/projects/dpl-cms/project-variables

I think I'd prefer if we could set the variables in the env-* repos instead of in the Lagoon system, because that feels like how we do things generally. But I haven't really succesfully passed in an env var here yet.

Here's a list of how to pass env vars from lagoon docs: https://docs.lagoon.sh/concepts-advanced/environment-variables/

From this list it looks like setting the environment variable in the Lagoon system is the right way to do it. c.f. `task lagoon:ensure:environment-variable` [here](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/main/infrastructure/Taskfile.yml#L528) and consider implementing setting the variable as a part of what happens in `task lagoon:project:set` where we ensure the state of a lagoon project. This is also where we [set Github credentials for each project](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/main/infrastructure/Taskfile.yml#L665) which ends up being very similar.

hypesystem (changes requested):
All the terraform code is \*chef's kiss\*

I don't think this is the right approach for passing in env vars. Added notes on why and a suggestion for a path that will work.

hypesystem (approved):

PR #285, opened 2024-04-30
Enable sending mails to contacts with the admin password and info about where to login

ITViking:
#### What does this PR do?
This PR adds some tasks, one for sending emails, one for generating passwords for the admin user on main environments and a higher level abstraction which couples the two former tasks together to sent emails to library contact persons. 

This should be read through. It can be easily tested by running either of the tasks. 
The format for the /contacts-and-their-sites.yaml can be seen inside it. Just change the sitename, name(s) and the email(s)
If the az cli complains. run these two commands one after the other: 
`az extension remove -n communication
az extension add -n communication -y
`
https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-60
https://reload.atlassian.net/browse/DDFDRIFT-85

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -1082,3 +1082,13 @@ tasks:
         msg: "Could not find directory {{.dir_infra}}"
       - sh: "[ -d {{.dir_configuration}} ]"
         msg: "Could not find directory {{.dir_configuration}}"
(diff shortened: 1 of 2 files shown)


hypesystem (approved):
LGTM

PR #303, opened 2024-05-03
Add support for `importTranslationsCron` in sites.yaml

hypesystem:
#### What does this PR do?

This adds support for configuring cron schedule for an individual site for importing translations, and sets it for cms-school, so we get hourly translation imports in that one environment.

This allows us to define custom schedules for when translations should be imported.

Verify that the generated [.lagoon.yml](https://github.com/danishpubliclibraries/env-cms-school/blob/main/.lagoon.yml#L66-L71) looks correct

[DDFDRIFT-101](https://reload.atlassian.net/browse/DDFDRIFT-101?atlOrigin=eyJpIjoiZWZkMWY5ZDI2YjgzNDM0ZjhiZTdkZGE1Zjc1NjRhZjAiLCJwIjoiaiJ9)

[DDFDRIFT-101]: https://reload.atlassian.net/browse/DDFDRIFT-101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/environments/dplplat01/sites.yaml
+++ b/infrastructure/environments/dplplat01/sites.yaml
@@ -24,6 +24,7 @@ sites:
     releaseImageRepository: ghcr.io/danskernesdigitalebibliotek
     releaseImageName: dpl-cms-source
     dpl-cms-release: "2024.18.3"
+    importTranslationsCron: "0 * * * *"
     deploy_key: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIJ6SzfPFf/XeLeqI342kxuJAlATpDMtgAfqlrLTTbW2m"
   customizable-canary:
     name: "Customizable bibliotek - eksempel"
--- a/infrastructure/dpladm/bin/dpladm-shared.source
+++ b/infrastructure/dpladm/bin/dpladm-shared.source
@@ -66,10 +66,12 @@ function renderProfileTemplate {
   local releaseImageRepository=$3
   # Name of the container-image that contains the built release
   local releaseImageName=$4
+  # Cron schedule for importing translations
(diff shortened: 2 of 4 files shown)


infrastructure/environments/dplplat01/sites.yaml
```diff
@@ -24,6 +24,7 @@ sites:
     releaseImageRepository: ghcr.io/danskernesdigitalebibliotek
     releaseImageName: dpl-cms-source
     dpl-cms-release: "2024.18.3"
+    importTranslationsCron: "0 * * * *"
```
spaceo, line 27: Maybe the default translation import cron schedules should be defined everywhere in the sites yaml file - also on the std sites.
I am aware of the defaults [here](https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/303/files#diff-52b177fa1883f37996c67a5555015c80688f680bdfe28462b76d38e98f6c81c3R120-R123). But having them defined in sites.yaml makes it more visible what the differences are in the settings - I think.

spaceo (commented):
Looks good have a comment for you to consider

infrastructure/environments/dplplat01/sites.yaml
hypesystem: It's a good point, and easy to implement (non-breaking) if we want to at some point in the future. I'll leave it for now but keep it in mind for future changes to sites.yaml.

As we start having regular releases on several different plans etc we'll learn what makes sense in sites.yaml, and how verbose we want to be :-)

spaceo (approved):
LGTM 👍

PR #306, opened 2024-05-03
Scale up ingress-nginx, based on metrics

hypesystem:
#### What does this PR do?

Based on data from Grafana this seems like a reasonable level to request to be able to deal with spikes in traffic.

The ingresses (15!) are currently using around 300-400 Mi of memory and barely perceptible CPU. We would like to run on slightly fewer ingresses with room to scale up, so this request level (1.1Gi) should accomodate that wish.

Memory graph:

Screen Shot 2024-05-03 at 17 23 56

CPU graph:

Screen Shot 2024-05-03 at 17 25 27

We have 16 Gb memory available on each node, and allowing an ingress to be a memory heavy instance makes sense. Even at this level, we could accomodate 14 ingresses on a node, which is a lot! So it seems reasonable.

The change has been executed and is running in the platform.

[DDFDRIFT-96](https://reload.atlassian.net/browse/DDFDRIFT-96?atlOrigin=eyJpIjoiMjY1Yzk4MTBhMTQ2NGVmZGEwNGUwYmE4ZTQyYWI5ZTEiLCJwIjoiaiJ9)

[DDFDRIFT-96]: https://reload.atlassian.net/browse/DDFDRIFT-96?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/environments/dplplat01/configuration/ingress-nginx/ingres-nginx-values.template.yaml
+++ b/infrastructure/environments/dplplat01/configuration/ingress-nginx/ingres-nginx-values.template.yaml
@@ -20,8 +20,8 @@ controller:
     maxReplicas: 15
   resources:
     requests:
-      cpu: 500m
-      memory: 500Mi
+      cpu: 300m
+      memory: 1100Mi
   service:
     loadBalancerIP: "${INGRESS_IP}"
     externalTrafficPolicy: "Local"


hypesystem:
Ligner at vi har ramt et godt punkt med denne: HPA'en har skaleret ned til 5, og vi er klar til spikes i load, let's go!

achton (approved):
LGTM 👍🏻

PR #308, opened 2024-05-06
Disable autogenerated routes if primary route set

hypesystem:
#### What does this PR do?

Does what it says on the tin

[DDFDRIFT-97](https://reload.atlassian.net/browse/DDFDRIFT-97)


ITViking (approved):
looks good

PR #311, opened 2024-05-06
Adjusted pod autoscaler for sites

hypesystem:
#### What does this PR do?

The base levels for nginx-php pods (requests) 110 Mi memory and 0.02 cpu (by adding the requests for the two containers). Our estimates for a "large" site is that it will use 400 Mi memory and 0.1 cpu. At that point it may make sense to scale, so we adjust levels to fit that - still a very reasonable and low level.

So we set targets for the autoscaler that match a "large" instance and start scaling at that point. This means upping the CPU scaling to 500%, which is 0.1 cpu. Even this may be a low point to start scaling at...

Judge the reasoning. Should we up the targets even more to avoid needless scaling? We want to be able to pack into relatively few nodes here when load is low.

[DDFDRIFT-96](https://reload.atlassian.net/browse/DDFDRIFT-96)

[DDFDRIFT-96]: https://reload.atlassian.net/browse/DDFDRIFT-96?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infrastructure/dpladm/autoscaler.template.yaml
+++ b/infrastructure/dpladm/autoscaler.template.yaml
@@ -15,7 +15,7 @@ spec:
   - resource:
       name: cpu
       target:
-        averageUtilization: 70
+        averageUtilization: 500
         type: Utilization
     type: Resource
   minReplicas: 1


achton (approved):
Let's give it a go. I think this is one of those changes that require some monitoring (of e.g. Herlev + Billund) to determine whether we are hitting autoscaling targets.

It's still a balance between allowing autoscaling during spikes and also packing pods onto few nodes during regular operation. So a bit of trial and error, I'd say.

hypesystem:
Agreed

PR #323, opened 2024-05-08
Add task for looping through sites and checking if they have a logo set up

hypesystem:
#### What does this PR do?

We use this to get an indication of how many sites have started using the system.

It has the nice added benefit of putting the frontpage in varnish's cache.

[DDFDRAFT-113](https://reload.atlassian.net/browse/DDFDRIFT-113)

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -1019,6 +1019,15 @@ tasks:
       preconditions:
       - *require_site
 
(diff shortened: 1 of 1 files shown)


ITViking:
kan jeg teste det her? Det burde jeg vel kunne uden at det påvirker noget?

ITViking (approved):
works = approved

PR #327, opened 2024-05-14
Deploy 2024.20.1 to Canary

kasperg:


hypesystem (approved):
In the future, please note on the PR whether or not you have run the change, so it is executed into the infrastructure.

I can also see that the actual [version in env-canary](https://github.com/danishpubliclibraries/env-canary/blob/main/lagoon/cli.dockerfile#L4) is 2024.20.2, so some code has been executed and not pushed to a PR :smile:

PR #329, opened 2024-05-15
patch for 2024.20.2

ITViking:
#### What does this PR do?
It patches 2024.20.2 and adds that as a repeatable image source we can roll out to specific sites. It also allows us to keep track of state without polluting sites.yaml


kasperg (approved):
:+1: LGTM

PR #341, opened 2024-05-27
Node pools fewer max pods

hypesystem:
#### What does this PR do?

Migrates to node pools with slightly lower max_pods setting. This should give workloads more breathing room to use more resources when needed.

We are limited by Lagoon hardcoding resource requests for pods. At 100 pods it seemed to push the limits of the nodes quite harshly, resulting in slow scheduling and some lowered performance.

The change has been executed in the platform.

[DDFDRIFT-128](https://reload.atlassian.net/browse/DDFDRIFT-128)

[DDFDRIFT-128]: https://reload.atlassian.net/browse/DDFDRIFT-128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


achton (approved):
LGTM 👍🏻

PR #349, opened 2024-05-30
Autogenerate redirect routes

hypesystem:
#### What does this PR do?

Allows a new value for `autogenerateRoutes` called `"redirect"` which indicates that we want traffic to the auto-generated routes (nginx.main.<site>... and varnish.main.<site>...) to be redirected to the primary domain set on the site.

Tested and executed on canary, c.f. https://nginx.main.canary.dplplat01.dpl.reload.dk/

Tested the equivalence with previous approach for aalborg, c.f. changes (only the order has changed): https://github.com/danishpubliclibraries/env-aalborg/commit/e883478b78216710c46c1f78b5bc5cae94c4d50a

The move of launched domains `autogenerateRoutes: {true->"redirect"}` has **not** been executed.

Read through and validate that this approach makes sense. Test the executed changes mentioned above.

[DDFDRIFT-138: Få domæner sat op for Struer og Faxe så de kan gå live når de vil](https://reload.atlassian.net/browse/DDFDRIFT-138)

### DIFF
--- a/infrastructure/dpladm/bin/dpladm-shared.source
+++ b/infrastructure/dpladm/bin/dpladm-shared.source
@@ -77,18 +77,31 @@ function renderProfileTemplate {
 
   PRIMARY_DOMAIN=""
 
+  if [[ "${autogenerateRoutes}" == "redirect" ]]; then
+    autogenerateRoutes="false"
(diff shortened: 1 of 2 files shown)


ITViking (approved):
Looks freaking good my man

PR #355, opened 2024-06-04
Webmaster pilot sites

ITViking:
#### What does this PR do?
This PR assigns København, Herning, Faxe, Tårnby, Århus and Aalborg to the webmaster plan. When run, this will result in them having `moduletest` environments. 

**This has only been run on canary and not the actual libraries** 

This should be read throuh, ensuring that the libraries mentioned above has the property `plan`set to `webmaster`

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-137


kasperg (approved):
:+1: I have gone through the list of sites and it corresponds to what has been agreed on.

Note that a subsequent deployment has been applied to canary without this being in in #356.

In other words: After deploying this to Canary, another deployment where the site was not marked as having the webmaster plan has been deployed. I do not know if this causes a problem.

hypesystem (approved):

PR #387, opened 2024-06-26
Add task to show the latest backup for each site

kasperg:
#### What does this PR do?

Currently we have no good way of identifying failing backups. Our best way to check that they are running for all sites is to check when the last succesfull backup was run.

To help in this process we add a new task which retrieves this from each site in Lagoon and prints it.

Use silent to avoid printing the shell command for each site. This makes the output more compact.

Try running the task for yourself.

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -1066,6 +1066,18 @@ tasks:
             | grep -v ": 0 matches"
           echo "done"
 
+    sites:latest-backup:
+      desc: Show the timestamp for the last successful backup for each site
+      vars:
+        SITES:
+          sh: lagoon list projects --output-json | jq -r ".data[].projectname"
+      cmds:
+        - for: { var: SITES }
+          cmd: |
+            DATE=$(lagoon list backups -p {{.ITEM}} -e main --output-json | jq -r '.data[0].created')
+            echo "{{.ITEM}}: ${DATE}"
+          silent: true
+
     site:lagoon:project:capture-deploy-key:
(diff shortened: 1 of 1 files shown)


infrastructure/Taskfile.yml
```diff
@@ -1066,6 +1066,18 @@ tasks:
             | grep -v ": 0 matches"
           echo "done"
 
+    sites:latest-backup:
+      desc: Show the timestamp for the last successful backup for each site
+      vars:
+        SITES:
+          sh: lagoon list projects --output-json | jq -r ".data[].projectname"
+      cmds:
+        - for: { var: SITES }
```
kasperg, line 1075: This is a new approach to running a task across all projects using [Taskfile loops](https://taskfile.dev/usage/#looping-over-values).

I think this reads a little better than our current approach with bash.

ITViking (approved):
Looks good and behaves as expected. Approved from here. A small suggestion, to more quickly assertain which projects haven't had a backup could be to see which projects doesn't have a backup in the last 14 hours.

kasperg:
It should be possible to fetch this in a single GraphQL query in the style of #392. We should consider this before doing this.

infrastructure/Taskfile.yml
hypesystem: 👍 we use it elsewhere, and it makes a lot of sense here

hypesystem (approved):

hypesystem:
I will merge this now, but if we do an improved graphql version later that's also fine :-)

PR #390, opened 2024-07-02
Switch to new nodepool

hypesystem:
#### What does this PR do?

Slowly switches to E-series boxes for our workloads. They are high in memory and lower in CPU, which fits our usage profile better for the money.

Also improved SLA for AKS cluster and database settings.

Now fully switched over. Has been executed.


ITViking (approved):
ser godt ud - go merge

PR #392, opened 2024-07-03
Refactor incomplete deployments task

kasperg:
#### What does this PR do?

Fetch deployment status for all projects in a single GraphQL query. This should be much more efficient compared to listing all projects and then retrieving the deployment status for each of them.

Try to run the task yourself.

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -1046,10 +1046,18 @@ tasks:
         Gets the latest deployment for each production environment and prints its status if it is *not* complete.
       cmds:
         - |
-          lagoon list projects --output-json \
-            | jq -r '.data[].projectname'  \
-            | while read -r projectname; do echo "$projectname: $(lagoon list deployments -e main -p $projectname --output-json | jq '.data[0].status')"; done \
-            | grep -v "complete"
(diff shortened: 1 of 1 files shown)


infrastructure/Taskfile.yml
```diff
@@ -1046,10 +1046,18 @@ tasks:
         Gets the latest deployment for each production environment and prints its status if it is *not* complete.
       cmds:
         - |
-          lagoon list projects --output-json \
-            | jq -r '.data[].projectname'  \
-            | while read -r projectname; do echo "$projectname: $(lagoon list deployments -e main -p $projectname --output-json | jq '.data[0].status')"; done \
-            | grep -v "complete"
+          lagoon raw --raw "query allProjects {
```
ITViking, line 1049: I'm getting an error when running this. It says that it doesn't know `raw`. It seems that `dplsh`'s cli is outdated? What version are you running?

kasperg: You are right.

I am runing `lagoon v0.21.3 (go-1.21)` locally but DPLSH has `lagoon v0.20.2 (go-1.16)`

ITViking (approved):
it works - LGTM

PR #403, opened 2024-07-11
Seperate versions between moduletest and main

ITViking:
#### What does this PR do?
This PR allows us to have seperate versions on WM and prod sites

Read it through and test it on canary 

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-154

### DIFF
--- a/infrastructure/dpladm/bin/sync-site.sh
+++ b/infrastructure/dpladm/bin/sync-site.sh
@@ -34,6 +34,19 @@ function getSiteDplCmsRelease {
     return
 }
 
+
+function getWebmasterDplCmsRelease {
+    local wmRelease
+    wmRelease=$(yq eval ".sites.${1}.webmaster-cms-version" "${2}")
+    if [[ "${wmRelease}" == "null" ]]; then
+        echo ""
+        return
+    fi
+
+    echo "${wmRelease}"
+    return
+}
+
 function getSiteReleaseImageRepository {
     local repository
     repository=$(yq eval ".sites.${1}.releaseImageRepository" "${2}")
--- a/infrastructure/dpladm/bin/sync-site.sh
+++ b/infrastructure/dpladm/bin/sync-site.sh
@@ -37,7 +37,7 @@ function getSiteDplCmsRelease {
 
 function getWebmasterDplCmsRelease {
     local wmRelease
-    wmRelease=$(yq eval ".sites.${1}.webmaster-cms-version" "${2}")
+    wmRelease=$(yq eval ".sites.${1}.moduletest-dpl-cms-release" "${2}")
     if [[ "${wmRelease}" == "null" ]]; then
         echo ""
         return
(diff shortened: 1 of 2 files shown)


infrastructure/environments/dplplat01/sites.yaml
```diff
@@ -38,6 +38,7 @@ sites:
     description: "Eksempel på bibliotek der kører på 'webmaster' plan, og derfor har et modultest-miljø"
     deploy_key: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAILj8lXv7C/7s7te9sEpwcHQhgDWfzsCkAN7rqQ4sdTzk"
     plan: webmaster
+    webmaster-cms-version: "2024.27.1"
```
kasperg, line 41: The other release related entry name is called `dpl-cms-release`. This is called `webmaster-cms-version` and is used for the `moduletest` environment.

I wonder if we can come up with something that shows a closer relation between the parts. Suggestion:

- `webmaster-dpl-cms-release` - the existing name prefixed with `webmaster` to show the relationship with the plan
- `moduletest-dpl-cms-release`- the existing name prefixed with the environment name
- `webmaster-moduletest-release` - both plan and environment name without the `dpl-cms` name

infrastructure/dpladm/bin/sync-site.sh
```diff
@@ -34,6 +34,19 @@ function getSiteDplCmsRelease {
...
+    local wmRelease
+    wmRelease=$(yq eval ".sites.${1}.webmaster-cms-version" "${2}")
+    if [[ "${wmRelease}" == "null" ]]; then
+        echo ""
+        return
+    fi
+
+    echo "${wmRelease}"
+    return
+}
```
kasperg, line 48: How will the caller `syncEnvRepo` handle an empty string as a response?

I am no expert in Bash but I wonder if it would be preferable if the function failed entirely if this is missing - or alternately used `releaseTag` with a message explaining that no information about a webmaster specific function could be found.

kasperg (changes requested):
:+1: I have some comments that I think at least should be considered before this is merged.

infrastructure/environments/dplplat01/sites.yaml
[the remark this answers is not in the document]
ITViking: done

infrastructure/dpladm/bin/sync-site.sh
ITViking: This is a copy of the already tried and tested function that finds the release for `main`. If there is an empty string that's what's returned. It then defaults to the latest image.

ITViking:
I've retested it. Main and moduletest of canary is now running seperate versions: 33.0 and 32.1 respectively

kasperg (approved):

PR #416, opened 2024-08-08
attemp to increase deployment speed

ITViking:
#### What does this PR do?
It attempts to increase the speed of deployments by allowing more consecutive builds to made

Just read it through

DDFDRIFT-184

### DIFF
--- a/infrastructure/environments/dplplat01/lagoon/lagoon-remote-values.template.yaml
+++ b/infrastructure/environments/dplplat01/lagoon/lagoon-remote-values.template.yaml
@@ -8,7 +8,7 @@ lagoon-build-deploy:
     - "--harbor-username=admin"
     - "--harbor-password=$HARBOR_ADMIN_PASS"
     - "--enable-qos"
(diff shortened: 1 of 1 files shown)


kasperg (approved):
I do not have great expectations about what results this will yield nor strong opinions regarding whether we should make this change or not. As such approved.

[We lowered the number of consecutive builds previously](https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/400). We do not have strong data on this but as I see it this did not have a significant positive impact regarding deployment stability nor a negative impact on deployment speed.

If we decide to make this change then I suggest that we keep a written record on speed and stability for future releases like we have done for previous releases like [`2024.29.0`](https://reload.zulipchat.com/#narrow/stream/355725-DDF---Drift/topic/Release.202024.2E29.2E0) and [`2024.28.0`](https://reload.zulipchat.com/#narrow/stream/355725-DDF---Drift/topic/Release.202024.2E28.2E0/near/450885256).

PR #420, opened 2024-08-13
DRAFT: Make it hard to deploy a release with out at least two sets of eye having seen the changes

ITViking:
#### What does this PR do?
This PR attempts to hinde the error from last week where I deployed from a non-main branch in an unclean state. To phrase it a little differently: Try to ensure that a version bump mr has been approved by someone on the team before it is released

This can be tested on canary for now with the command `site:sync SITE=canary`. It should fail as you're not on the main branch. 
You should also test it by changing branch name from `main` to `prevent-unsafe-releases` and making a change somewhere. This should also be caught. 

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-183

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -897,6 +897,26 @@ tasks:
           fi
       preconditions:
       - *require_site
+      - &require_clean_state {
+          sh: "[
+            current_branch=$(git branch --show-current)
+            if [ 'current_branch' != 'main' ]; then
+              echo 'Can't deploy if not on main branch'
+              exit 1
+            fi
+            if ! git diff-index --quiet HEAD --; then
+              echo 'Can't deploy if there's changes on the branch'
+              exit 1
+            fi
+            exit 0 #
+          ]",
+          msg: "############################################################ \n
+                \n
+                It's not possible to deploy from other branches than main.\n
+                Furthermore your state has to be clean \n
+                \n
+                ############################################################"
+        }
 
     site:autoscaler:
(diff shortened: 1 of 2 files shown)


infrastructure/Taskfile.yml
```diff
@@ -897,6 +897,26 @@ tasks:
...
+            if [ 'current_branch' != 'main' ]; then
+              echo 'Can't deploy if not on main branch'
+              exit 1
+            fi
+            if ! git diff-index --quiet HEAD --; then
+              echo 'Can't deploy if there's changes on the branch'
+              exit 1
+            fi
+            exit 0
+          ]",
```
kasperg, line 912: This checks that all local changes have been committed and merged to the main branch.

Consider whether we can (or should) also require the changes to be pushed to GitHub.

infrastructure/Taskfile.yml
```diff
@@ -897,6 +897,26 @@ tasks:
           fi
       preconditions:
       - *require_site
+      - &require_clean_state {
```
spaceo, line 900: I am really not sure aboyt the [] encapsulation, but if you say it works as a multiline delimiter I trust you :)
I think there are problems with quoting and a small typo:
```bash
  current_branch=$(git branch --show-current)
  if [ "$current_branch" != "main" ]; then
    echo "Can't deploy if not on main branch"
    exit 1
  fi

  if ! git diff-index --quiet HEAD --; then
    echo "Cant deploy if there are changes on the branch"
    exit 1
  fi

  exit 0

spaceo (changes requested):
Some comments...

PR #424, opened 2024-08-15
Temporary script for handling I/O errors

achton:
#### What does this PR do?

This is a script which will iterate over all `php` containers in nginx pods in every namespace, check if `df` reports and I/O error, and restart the deployment if so.

The PR should probably be closed once a permanent fix for the I/O error is in place.


ITViking (approved):
LGTM - it deserves it :)

ITViking:
This is no longer needed

PR #431, opened 2024-08-23
up the number of nodes

ITViking:
#### What does this PR do?
Does it says on the tin

DDFDRIFT-196


ITViking:
We are out of sync with terraform, so this makes no sense anymore

PR #434, opened 2024-08-26
Redirect legacy paths for library search and materials to new paths

kasperg:
#### What does this PR do?

This is a mirror of danskernesdigitalebibliotek/dpl-cms#1508 which applies the change for the NGINX image to the environment template.

If accepted then we must ensure that the changes to the template are propagated to the sites.

Consider checking the test in the danskernesdigitalebibliotek/dpl-cms#1508

https://reload.atlassian.net/browse/DDFFORM-978


spaceo (approved):
LGTM 👍 

👍  for remembering to do it both in dpl-cms an dpl-platform 😅

PR #440, opened 2024-08-30
Check that arguments is defined in syncEnvRepo

xendk:
#### What does this PR do?

Add some arguments checking to syncEnvRepo so it won't work with a missing tag, for instance.

Nope, I'm not entirely sure of the environment to run it in. But the `if` syntax has been checked with shellcheck.


ITViking (approved):
good improvement - LGTM

PR #441, opened 2024-09-02
set version for moduletest

ITViking:
#### What does this PR do?
set the moduletest environments version

Just read it

No

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-200

### DIFF
--- a/infrastructure/environments/dplplat01/sites.yaml
+++ b/infrastructure/environments/dplplat01/sites.yaml
@@ -70,6 +70,7 @@ sites:
     description: "The library site for Aarhus"
     deploy_key: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIFL+uMeEfsaHEzbNxOmBB8dX32OLo63CTomG8VZvuiN2"
     plan: webmaster
+    moduletest-dpl-cms-release: "2024.35.1"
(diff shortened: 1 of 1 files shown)


xendk (approved):
Look good.

But why don't we parameterize moduletest so we only have to change it at one place?

infrastructure/environments/dplplat01/sites.yaml
```diff
@@ -70,6 +70,7 @@ sites:
     description: "The library site for Aarhus"
     deploy_key: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIFL+uMeEfsaHEzbNxOmBB8dX32OLo63CTomG8VZvuiN2"
     plan: webmaster
+    moduletest-dpl-cms-release: "2024.35.1"
```
spaceo, line 73: Since the moduletest-dpl-cms-release version equals the default version why not using only the default version and delete the moduletest-dpl-cms-release version until it differs in the future?

ITViking: you're right. Last week i had to do it a little fast, so that's why i did it like this

PR #442, opened 2024-09-02
Disable caching of static files in Varnish

kasperg:
#### Link

https://reload.atlassian.net/browse/DDFDRIFT-216

#### What does this PR do?

This is a mirror of danskernesdigitalebibliotek/dpl-cms#1522 which applies the change for the NGINX image to the environment template.

If accepted then we must ensure that the changes to the template are propagated to the sites.

Consider checking the test in the danskernesdigitalebibliotek/dpl-cms#1522


ITViking (approved):
Det ser godt ud - jeg har kigget i PR miljøet og det ser også fint ud. Lad os se om ikke det hjælper

PR #448, opened 2024-09-06
Switch from in tree to csi

ITViking:
#### What does this PR do?
These are the scripts we have used to move to the new CSI

no

read it 

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?search=csi&selectedIssue=DDFDRIFT-189


ITViking:
@ath88 I'm considering closing this for now as we will probably not need it again. If we do, we know where to find it - thoughts?

ITViking:
Closing this, as we are hopefully not going to need this in the future. If we do, we know where it is

PR #455, opened 2024-09-16
Use latest version on some webmaster sites

kasperg:
#### What does this PR do?

[Add an anchor for webmaster sites using the latest version on main](https://github.com/danskernesdigitalebibliotek/dpl-platform/commit/89bce9345c8fb3204bd5c4b35e0e6a6b4c6162c5)

Normally webmaster sites use a previous version on their main
(production) environment and the latest version on their moduletest
environment.

However some sites want to use the latest version on both
environments.

Create an anchor to use for these sites.

Reference the new anchor for CMS School and Silkeborg as requested by
the business.

Reference the new anchor on staging to ensure that this follows the
latest version. This makes sense to enable testing there.

Deploy the changes to the three sites in question.

One has to understand the intricacies of mulitple YAML references, overrides etc. to understand how the new anchor works.
This is not optimalt but I fear that we will forget to update these sites otherwise in the future.

### DIFF
--- a/infrastructure/environments/dplplat01/sites.yaml
+++ b/infrastructure/environments/dplplat01/sites.yaml
@@ -8,6 +8,13 @@ x-webmaster: &webmaster-release-image-source
   releaseImageName: dpl-cms-source
   dpl-cms-release: "2024.35.1"
   moduletest-dpl-cms-release: "2024.37.0"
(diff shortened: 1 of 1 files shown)


ITViking (approved):
This looks well - let's go

ITViking (approved):

PR #461, opened 2024-09-18
Add tolerations to cert manager and fix promtail

ath88:
#### What does this PR do?
- We add tolerations and affinities to cert manager workloads 
- fix promtail by letting it deploy daemonset on nodes with specific taints

No

No

DDFDRIFT-86


ITViking (approved):
looks good - lets go :D

PR #462, opened 2024-09-18
Set tolerations and affinity for prometheus

ath88:
#### What does this PR do?
Adds tolerations and affinity for prometheus workloads

86


ITViking (approved):
also very fine

PR #486, opened 2024-10-09
Task for cluster deploy mode

ITViking:
#### What does this PR do?
It adds the ability to set the cluster into deploy mode and back into business mode

Please run the both tasks on after the other.
After having set the cluster in deploy mode, go here: https://portal.azure.com/#@reloadas.onmicrosoft.com/resource/subscriptions/8ac8a259-5bb3-4799-bd1e-455145b12550/resourceGroups/rg-env-dplplat01/providers/Microsoft.ContainerService/managedClusters/aks-dplplat01-01/nodePools
The app6 nodepool should now have 4 nodes and the system2 nodepool should have 5. 

Then set the cluster back into business mode and check that, the app6 pool has 2 nodes and the system2 pool has 3 nodes. 

Run the tasks and read it through

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-240

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -238,9 +238,9 @@ tasks:
           --resource-group rg-env-dplplat01
           --node-count 5
 
-    cluster:mode:business:
+    cluster:mode:reset:
       deps: [cluster:auth , lagoon:cli:config]
-      desc: "Set the cluster in release mode by adding additional nodes to release critical node pools"
+      desc: "Reset changes to the cluster mode returning it to standard operations."
       cmds:
         - az aks nodepool scale
           --cluster-name aks-dplplat01-01
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -235,7 +235,10 @@ tasks:
         - az aks nodepool scale
           --cluster-name aks-dplplat01-01
           --nodepool-name system2
-          --resource-group rg-env-dplplat01
+          --resource-group
+            $(
+              terraform -chdir={{.dir_infra}} output -json | jq --raw-output ".resourcegroup_name.value | select (.!=null)"
+            )
           --node-count 5
 
     cluster:mode:reset:
@@ -250,7 +253,10 @@ tasks:
         - az aks nodepool scale
           --cluster-name aks-dplplat01-01
           --nodepool-name system2
-          --resource-group rg-env-dplplat01
+          --resource-group
+            $(
+              terraform -chdir={{.dir_infra}} output -json | jq --raw-output ".resourcegroup_name.value | select (.!=null)"
+            )
           --node-count 3
 
     support:provision:cert-manager:
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -223,6 +223,21 @@ tasks:
       cmds:
         - task/scripts/promote-workloads-to-prod-nodes.sh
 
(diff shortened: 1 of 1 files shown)


infrastructure/Taskfile.yml
```diff
@@ -223,6 +223,36 @@ tasks:
...
+          --node-count 4
+        - az aks nodepool scale
+          --cluster-name aks-dplplat01-01
+          --nodepool-name system2
+          --resource-group rg-env-dplplat01
+          --node-count 5
+
+    cluster:mode:business:
+      deps: [cluster:auth , lagoon:cli:config]
+      desc: "Set the cluster in release mode by adding additional nodes to release critical node pools"
```
kasperg, line 243:
```suggestion
    cluster:mode:reset:
      deps: [cluster:auth , lagoon:cli:config]
      desc: "Reset changes to the cluster mode returning it to standard operations."
```

The description looks like copy/paste and should be changed :smile:.

Changing the name is just a suggestion.

infrastructure/Taskfile.yml
```diff
@@ -223,6 +223,36 @@ tasks:
...
+        - az aks nodepool scale
+          --cluster-name aks-dplplat01-01
+          --nodepool-name app6
+          --resource-group rg-env-dplplat01
+          --node-count 2
+        - az aks nodepool scale
+          --cluster-name aks-dplplat01-01
+          --nodepool-name system2
+          --resource-group rg-env-dplplat01
+          --node-count 3
```
kasperg, line 254: In keeping with the `reset` name I wondered if node counts in nodepools/resource counts was something that had a default value in Terraform or the like?

If that is the case then I suggest then I wonder if we could run Terraform to reset the cluster to a default state and thus node count.

If not the move along :smile:

infrastructure/Taskfile.yml
```diff
@@ -223,6 +223,36 @@ tasks:
...
+      cmds:
+        - az aks nodepool scale
+          --cluster-name aks-dplplat01-01
+          --nodepool-name app6
+          --resource-group rg-env-dplplat01
+          --node-count 4
+        - az aks nodepool scale
+          --cluster-name aks-dplplat01-01
+          --nodepool-name system2
+          --resource-group rg-env-dplplat01
```
kasperg, line 238: Can we move some of these hard coded values out of the task?

I did a quick search through the codebase and as far as I can tell we do not have values such as `aks-dplplat01-01`,  `app6`, `system2` or `rg-env-dplplat01` hardcoded elsewhere.

kasperg (commented):
:+1: I appreciate the intention here.

I have some questions. Bear in mind my limited experience here :smile:

infrastructure/Taskfile.yml
ITViking: Thats a good suggestion - I was hoping someone would come up with a suggestion :)

infrastructure/Taskfile.yml
ITViking: Terraform is out of sync right now, otherwise that would've been the way to go about it

infrastructure/Taskfile.yml
ITViking: There no values at the moment for the cluster and nodepools as TF is out for sync, additionally the way this is set up, we're missing logic for choosing the right cluster and nodepools in the cluster

ath88 (approved):

PR #498, opened 2024-10-25
fix that autogenerated routes were not created when creating the new …

ITViking:
…routes

#### What does this PR do?
This ensures that autogenrated routes are also made


ath88:
What is "autogenerated routes" and why do we need it? And why doesn't all sites have it?

ath88 (approved):

PR #504, opened 2024-11-01
add task to synchonize environments

ITViking:
#### What does this PR do?
Add a task that given source and target can synchronize files and datbases in on go

Please do

Read and test it

DDFDRIFT-247


infrastructure/Taskfile.yml
```diff
@@ -1085,6 +1085,21 @@ tasks:
...
+    site:environment:sync:
+      desc: |
+        Synchonizes all file and database tables from environment
+        to another on make on environment a complete copy of the source.
+      deps: [lagoon:cli:config]
+      dir: "{{.dir_env}}"
+      vars:
+        SOURCE_PROJECT: "{{.SOURCE_PROJECT}}"
+        SOURCE_ENV: "{{.SOURCE_ENV}}"
+        TARGET_PROJECT: "{{.TARGET_PROJECT}}"
```
kasperg, line 1097: Does this approach work across projects?

I wonder if `drush rsync` has the necessary configuration availabele to synchronize between different projects. In any case we might consider only allowing a single `PROJECT` variable.

infrastructure/Taskfile.yml
```diff
@@ -1085,6 +1085,21 @@ tasks:
...
+        TARGET_ENV: "{{.TARGET_ENV}}"
+      cmds:
+        - lagoon ssh --project {{.TARGET_PROJECT}} --environment {{.TARGET_ENV}} -C "drush -y rsync @lagoon.{{.SOURCE_PROJECT}}-{{.SOURCE_ENV}}:%files @self:%files -- --omit-dir-times --no-perms --no-group --no-owner --no-times --chmod=ugo=rwX --delete; drush -y sql-sync @lagoon.{{.SOURCE_PROJECT}}-{{.SOURCE_ENV}} @self; exit;"
```
kasperg, line 1100: :+1: I see that you added the `--delete` argument to `drush rsync`. Very nice!

With that in mind: Could you also add the `--create-db` option to `drush sql-sync`? This way we avoid any additional database tables added to the `moduletest` environment from lingering after a sync.

[We have discussed this in Zulip](https://reload.zulipchat.com/#narrow/channel/419623-DDF.2B/topic/Webmasterl.C3.B8sning/near/478069316).

kasperg, line 1100: If you are interested in queuing this task and you can consider [running it as a Lagoon custom task](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/f464992fcc50547b26180c3c350618bbba8566f4/task/Taskfile.lagoon.yml#L50-L50). 

This requires that the subsequent deploy is queued to be run after the task and not executed immediatly.

kasperg, line 1100: As I see it some of these commands are based on the implementation in Lagoon. Consider adding references to these as code comments.

kasperg (approved):
:+1: Nice. I have a few suggestions for you to consider before you merge.

infrastructure/Taskfile.yml
ITViking: I have not tested wether this works across projects. It is not intended for that right now

infrastructure/Taskfile.yml
ITViking: I don't think I'm quite following you here, please elaborate

ITViking: fixed

PR #509, opened 2024-11-04
Task to reset all moduletests

ITViking:
#### What does this PR do?
It finds all webmasters and resets all their moduletest environments by copying their files and databases from prod to moduletest and then running another deploy. 

Don't run the task unless it needs to. I have a tested it on the a list of canaries. 
Just read it

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-251

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -1099,6 +1099,16 @@ tasks:
       cmds:
         - lagoon ssh --project {{.TARGET_PROJECT}} --environment {{.TARGET_ENV}} -C "drush -y rsync @lagoon.{{.SOURCE_PROJECT}}-{{.SOURCE_ENV}}:%files @self:%files -- --omit-dir-times --no-perms --no-group --no-owner --no-times --chmod=ugo=rwX --delete; drush -y sql-sync @lagoon.{{.SOURCE_PROJECT}}-{{.SOURCE_ENV}} @self; drush cr -y"
 
+    sites:webmaster:reset-moduletests:
+      desc: Reset all webmaster moduletests. It is possible to exempt some.
+      deps: [lagoon:cli:config]
+      vars:
+        webmasters:
+          sh: yq '... comments="" | .sites | with_entries(select(.value | has("plan"))) | keys | .[]' {{.dir_env}}/sites.yaml
+      cmds:
+        - for: { var: webmasters }
+          cmd: echo "{{.ITEM}}"
+
     sites:check-caa:
       desc: |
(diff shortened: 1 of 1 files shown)


infrastructure/Taskfile.yml
```diff
@@ -1085,6 +1085,35 @@ tasks:
...
+      vars:
+        SOURCE_PROJECT: "{{.SOURCE_PROJECT}}"
+        SOURCE_ENV: "{{.SOURCE_ENV}}"
+        TARGET_PROJECT: "{{.TARGET_PROJECT}}"
+        TARGET_ENV: "{{.TARGET_ENV}}"
+      cmds:
+        - lagoon ssh --project {{.TARGET_PROJECT}} --environment {{.TARGET_ENV}} -C "drush -y rsync @lagoon.{{.SOURCE_PROJECT}}-{{.SOURCE_ENV}}:%files @self:%files -- --omit-dir-times --no-perms --no-group --no-owner --no-times --chmod=ugo=rwX --delete; drush -y sql-sync @lagoon.{{.SOURCE_PROJECT}}-{{.SOURCE_ENV}} @self; drush cr -y"
+
+    sites:webmaster:reset-moduletests:
+      desc: Reset all webmaster moduletests. It is possible to exempt some.
```
kasperg, line 1103: I do not understand how it is possible to exempt a site based on the implementation. Am I missing something?

infrastructure/Taskfile.yml
```diff
@@ -1085,6 +1085,35 @@ tasks:
...
+      deps: [lagoon:cli:config]
+      vars:
+        webmasters:
+          sh: yq '... comments="" | .sites | with_entries(select(.value | has("plan"))) | keys | .[]' {{.dir_env}}/sites.yaml
+      cmds:
+        - for: { var: webmasters }
```
kasperg, line 1109: :+1: This is very nice usage of Taskfile.

As I see it each invocation is blocking. You could consider queing them up as tasks instead as suggested in #504. This then creates the challenge of monitoring the progress of each sync 😕 .

kasperg (approved):
:+1: Based solely on the new commits here.

infrastructure/Taskfile.yml
ITViking: Then maybe we shouldn't queue them up so we can keep an eye on which fail to be queued :)

infrastructure/Taskfile.yml
ITViking: fixed - good point

PR #511, opened 2024-11-05
Task to restart failed deployements that are not from the dpl-cms project

ITViking:
#### What does this PR do?
This is hopefully a little nifty task that will make it faster and easier for us to redeploy failed deployments for both main and moduletest environments. 
I have done this as most deployment problems are solved by runnning them again. 
We should still investigate why deployments fail.

Just run it

Read it and test it 

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-252

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -1146,7 +1146,7 @@ tasks:
 
     sites:redeploy-failed-deployments:
       deps: [lagoon:cli:config]
-      desc: redeployyes all failed deployments
+      desc: Redeploys all failed deployments
       cmds:
         - task/scripts/redeploy-failed-deployments.sh
 
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -1134,7 +1134,7 @@ tasks:
           lagoon raw --raw "query allProjects {
             allProjects {
             name
-              environments(type: PRODUCTION) {
+              environments(type: DEVELOPMENT) {
                 name
                 deployments(limit: 1) {
                   status
@@ -1143,6 +1143,12 @@ tasks:
               }
             }
           }" | jq -r '.allProjects[] | .name as $name | .environments[].deployments[] | select(.status != "complete") | ($name) + ": " + (.status)' 
+
+    sites:redeploy-failed-deployments:
+      deps: [lagoon:cli:config]
+      desc: redeployyes all failed deployments
+      cmds:
+        - task/scripts/redeploy-failed-deployments.sh
 
     sites:grep-in-deploy-log:
       desc: |
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -1134,7 +1134,7 @@ tasks:
           lagoon raw --raw "query allProjects {
             allProjects {
             name
-              environments(type: DEVELOPMENT) {
+              environments(type: PRODUCTION) {
                 name
                 deployments(limit: 1) {
                   status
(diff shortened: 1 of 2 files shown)


infrastructure/Taskfile.yml
```diff
@@ -1104,7 +1104,7 @@ tasks:
           lagoon raw --raw "query allProjects {
             allProjects {
             name
-              environments(type: PRODUCTION) {
+              environments(type: DEVELOPMENT) {
```
kasperg, line 1107: Is this change intentional?

The comment for the task still states "Gets the latest deployment for each *production* environment".

In general I also think it is more valuable for us to see deployments to production over development.

infrastructure/Taskfile.yml
```diff
@@ -1114,6 +1114,12 @@ tasks:
             }
           }" | jq -r '.allProjects[] | .name as $name | .environments[].deployments[] | select(.status != "complete") | ($name) + ": " + (.status)' 
 
+    sites:redeploy-failed-deployments:
+      deps: [lagoon:cli:config]
+      desc: redeployyes all failed deployments
```
kasperg, line 1119:
```suggestion
      desc: Redeploys all failed deployments
```

infrastructure/task/scripts/redeploy-failed-deployments.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+#
+# 80% of failed deployments is fixed by redeploying the failed deployment.
+# This script will help make that easier to do.
+
+echo "Will now start to redeployg failed deployments"
```
kasperg, line 6:
```suggestion
echo "Will now start to redeploying failed deployments"
```

infrastructure/task/scripts/redeploy-failed-deployments.sh
```diff
@@ -0,0 +1,57 @@
...
+      name
+      deployments(limit: 1) {
+        status
+        created
+      }
+    }
+  }
+}" | jq -r '.allProjects[] | .name as $name | .environments[].deployments[] | select(.status == "failed") | ($name)')
+
+echo "redeploying failed production environments"
```
kasperg, line 22:
```suggestion
echo "Redeploying failed production environments"
```

infrastructure/task/scripts/redeploy-failed-deployments.sh
```diff
@@ -0,0 +1,57 @@
...
+      name
+      deployments(limit: 1) {
+        status
+        created
+      }
+    }
+  }
+}" | jq -r '.allProjects[] | .name as $name | .environments[].deployments[] | select(.status == "failed") | ($name)')
+
+echo "redeploying failed moduletest environments"
```
kasperg, line 47:
```suggestion
echo "Redeploying failed moduletest environments"
```

infrastructure/task/scripts/redeploy-failed-deployments.sh
```diff
@@ -0,0 +1,57 @@
...
+
+for deployment in $FAILED_DEVELOPMENT_DEPLOYMENTS; do
+  if [[ $deployment =~ "dpl-cms" ]]; then
+    continue
+  fi
+  echo "$deployment: deploying"
+  lagoon deploy latest -p "$deployment" -e "moduletest" --force
+done
```
kasperg, line 55: There is some repetition here that I wonder if we could refactor to reusable functions. I do not think it is that bad though. Keeping it simple is also nice.

kasperg (approved):
:+1: Nice.

I think you need to reconsider the change to `sites:incomplete-deployments` but I trust you can do this on your own.

I also have a few other suggestions and comments for you to consider.

Consider updating the deployment runbooks to refer to this task.

infrastructure/task/scripts/redeploy-failed-deployments.sh
ITViking: I considered that too when I wrote it. My conclusion was that this is a temporary solution to a problem we want to solve more completely, so for now let's just keep this simple

PR #523, opened 2024-11-20
dplsh upgrades - nov 2024

achton:
#### What does this PR do?

This will upgrade some packages (Task, Stern and the Lagoon CLI) to latest releases, as `dplsh` was lagging behind somewhat. 

I've tested this against the main cluster with a locally built shell without issues.


ITViking (approved):
LGTM

PR #524, opened 2024-11-20
Increase ingress-nginx client_header_buffer_size

achton:
#### What does this PR do?

This moves the ingress-nginx config away from the [default Nginx value of 1kb for `client_header_buffer_size`](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_header_buffer_size), and sets our own value of 8kb for that setting instead.

The reason for this change is that some incoming requests to the application containers are erroring with a message like this:

`2024/11/20 09:58:39 [error] 60431#60431: *30679622 upstream sent too big header while reading response header from upstream, client: 5.103.135.109, server: node.pr-1757.dpl-cms.dplplat01.dpl.reload.dk, request: "GET /auth/callback/unilogin?session_state=111b7164-1637-4836-87fc-fa6f4e957c53&code=5ad2f288-c3eb-4b95-b501-3430173fdf8a.111b7164-1637-4836-87fc-fa6f4e957c53.eb9dd6e6-07f1-41e2-9ad9-b777d9ce69af HTTP/2.0", upstream: "http://10.1.12.195:3000/auth/callback/unilogin?session_state=111b7164-1637-4836-87fc-fa6f4e957c53&code=5ad2f288-c3eb-4b95-b501-3430173fdf8a.111b7164-1637-4836-87fc-fa6f4e957c53.eb9dd6e6-07f1-41e2-9ad9-b777d9ce69af", host: "node.pr-1757.dpl-cms.dplplat01.dpl.reload.dk"`

This is from Grafana in the `ingress-nginx` namespace, and the code behind this works on the developer's local machine, so this tells me that the ingress controller config is the culprit here.

This needs to be rolled out to production, and cannot be tested locally (AFAIK).

DDFBRA-217

### DIFF
--- a/infrastructure/environments/dplplat01/configuration/ingress-nginx/ingres-nginx-values.template.yaml
+++ b/infrastructure/environments/dplplat01/configuration/ingress-nginx/ingres-nginx-values.template.yaml
@@ -13,6 +13,10 @@ controller:
     # See https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size
     # and https://github.com/uselagoon/lagoon-images/blob/main/images/nginx/nginx.conf#L81
     proxy-body-size: "2048m"
+    # Default client_header_buffer_size is 1k in nginx, which results in 502
+    # errors for some applications, eg. large cookie headers.
+    # See https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#client-header-buffer-size
+    client-header-buffer-size: "8k"
   replicaCount: 2
   autoscaling:
     enabled: true


infrastructure/environments/dplplat01/configuration/ingress-nginx/ingres-nginx-values.template.yaml
```diff
@@ -13,6 +13,10 @@ controller:
     # See https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size
     # and https://github.com/uselagoon/lagoon-images/blob/main/images/nginx/nginx.conf#L81
     proxy-body-size: "2048m"
+    # Default client_header_buffer_size is 1k in nginx, which results in 502
+    # errors for some applications, eg. large cookie headers.
+    # See https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#client-header-buffer-size
+    client-header-buffer-size: "8k"
```
spaceo, line 19: While I do not doubt this might do the job. Are you sure that it is the only setting needed?
According to this post, a guy managed to fix the problem with some other settings:
https://community.auth0.com/t/next-js-app-returns-502-bad-gateway-for-api-auth-me-after-a-successful-login/135909/3

spaceo (commented):
@achton I have a question I would like to have clarified

infrastructure/environments/dplplat01/configuration/ingress-nginx/ingres-nginx-values.template.yaml
spaceo: Same story here:
https://stackoverflow.com/questions/57294212/502-bad-gateway-error-upstream-sent-too-big-header-while-reading-response-head

spaceo: And here:
https://unix.stackexchange.com/a/605614

achton: I'm not completely sure, but I think the `proxy_*` headers are only relevant with `proxy_pass` traffic, ie. when speaking with PHP-FPM.

In this case, we are trying to affect the HTTP header of an incoming client request, which is why I think [`client_header_buffer_size`](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#client-header-buffer-size) is most likely.

spaceo: Two of the links I posted refers to a setup with nginx in front of a running node application. Eg. [this](https://unix.stackexchange.com/questions/605467/how-to-handle-a-too-large-cookie-causing-nginx-to-return-a-502) one.
Wouldn't that always be in a proxy relation like the one we have with PHP-FPM?

achton: Possibly. This is the generated block of config from `nginx.conf` in a ingress-nginx container:

```lua

	## start server node.pr-1757.dpl-cms.dplplat01.dpl.reload.dk
	server {
		server_name node.pr-1757.dpl-cms.dplplat01.dpl.reload.dk ;
		
		listen 80  ;
		listen [::]:80  ;
		listen 443  ssl http2 ;
		listen [::]:443  ssl http2 ;
		
		set $proxy_upstream_name "-";
		
		ssl_certificate_by_lua_block {
			certificate.call()
		}
		
		# Custom code snippet configured for host node.pr-1757.dpl-cms.dplplat01.dpl.reload.dk
		add_header X-Robots-Tag "noindex, nofollow";
		
		location / {
			
			set $namespace      "dpl-cms-pr-1757";
			set $ingress_name   "node";
			set $service_name   "node";
			set $service_port   "http";
			set $location_path  "/";
			set $global_rate_limit_exceeding n;
			
			rewrite_by_lua_block {
				lua_ingress.rewrite({
					force_ssl_redirect = false,
					ssl_redirect = false,
					force_no_ssl_redirect = false,
					preserve_trailing_slash = false,
					use_port_in_redirects = false,
					global_throttle = { namespace = "", limit = 0, window_size = 0, key = { }, ignored_cidrs = { } },
				})
				balancer.rewrite()
				plugins.run()
			}
			
			# be careful with `access_by_lua_block` and `satisfy any` directives as satisfy any
			# will always succeed when there's `access_by_lua_block` that does not have any lua code doing `ngx.exit(ngx.DECLINED)`
			# other authentication method such as basic auth or external auth useless - all requests will be allowed.
			#access_by_lua_block {
			#}
			
			header_filter_by_lua_block {
				lua_ingress.header()
				plugins.run()
			}
			
			body_filter_by_lua_block {
				plugins.run()
			}
			
			log_by_lua_block {
				balancer.log()
				
				monitor.call()
				
				plugins.run()
			}
			
			port_in_redirect off;
			
			set $balancer_ewma_score -1;
			set $proxy_upstream_name "dpl-cms-pr-1757-node-http";
			set $proxy_host          $proxy_upstream_name;
			set $pass_access_scheme  $scheme;
			
			set $pass_server_port    $server_port;
			
			set $best_http_host      $http_host;
			set $pass_port           $pass_server_port;
			
			set $proxy_alternative_upstream_name "";
			
			client_max_body_size                    2048m;
			
			proxy_set_header Host                   $best_http_host;
			
			# Pass the extracted client certificate to the backend
			
			# Allow websocket connections
			proxy_set_header                        Upgrade           $http_upgrade;
			
			proxy_set_header                        Connection        $connection_upgrade;
			
			proxy_set_header X-Request-ID           $req_id;
			proxy_set_header X-Real-IP              $remote_addr;
			
			proxy_set_header X-Forwarded-For        $remote_addr;
			
			proxy_set_header X-Forwarded-Host       $best_http_host;
			proxy_set_header X-Forwarded-Port       $pass_port;
			proxy_set_header X-Forwarded-Proto      $pass_access_scheme;
			proxy_set_header X-Forwarded-Scheme     $pass_access_scheme;
			
			proxy_set_header X-Scheme               $pass_access_scheme;
			
			# Pass the original X-Forwarded-For
			proxy_set_header X-Original-Forwarded-For $http_x_forwarded_for;
			
			# mitigate HTTPoxy Vulnerability
			# https://www.nginx.com/blog/mitigating-the-httpoxy-vulnerability-with-nginx/
			proxy_set_header Proxy                  "";
			
			# Custom headers to proxied server
			
			proxy_connect_timeout                   5s;
			proxy_send_timeout                      60s;
			proxy_read_timeout                      60s;
			
			proxy_buffering                         off;
			proxy_buffer_size                       4k;
			proxy_buffers                           4 4k;
			
			proxy_max_temp_file_size                1024m;
			
			proxy_request_buffering                 on;
			proxy_http_version                      1.1;
			
			proxy_cookie_domain                     off;
			proxy_cookie_path                       off;
			
			# In case of errors try the next upstream server before returning an error
			proxy_next_upstream                     error timeout;
			proxy_next_upstream_timeout             0;
			proxy_next_upstream_tries               3;
			
			proxy_pass http://upstream_balancer;
			
			proxy_redirect                          off;
			
		}
		
	}
```

So some of the proxy directives are already being configured even if we do not set them explicitly in the ConfigMap.

spaceo: According to [this](https://unix.stackexchange.com/a/605614) post, we should increase: proxy_buffer_size and proxy_buffers. Do we have access to manipulate those values?

ITViking (approved):
As I see it, this is not to bad to test out. I think it is worth finding if this solves the problem, and if it does, then we should maybe try and see if we can make the size smaller, to fit the headers we are going to use more adequately. 
LGTM

ath88:
This setting is recommended to be 1k. In some situations, headers may be larger, but changing the default ingress setting will change it for every situation.

I would recommend you look into using the [server-snippet-annotation](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#server-snippet) on the specific ingress instead, as that will have a smaller impact. Something like this should do it:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    nginx.ingress.kubernetes.io/server-snippet: |
        client_header_buffer_size 8k;
...
```

Server snippets needs to be enabled in the ingress controller using `controller.enableSnippets: true`, so make sure that is enabled before testing.

Besides that, it sounds like `client_header_buffer_size` is the correct parameter to adjust. :)

achton:
@ath88 I know what you mean, but where to configure annotations for the `node` container and its ingress? I don't know where to find that configuration..

ath88:
How is the node application deployed?

spaceo:
> How is the node application deployed?

By adding a service block to docker-compose.yml:
https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/develop/docker-compose.yml#L144-L148

So this is running in the PR env's as it is now. Has be a part of the site template as well when it needs to hit production.

achton:
Okay, well my question then becomes: how do we configure the ingress resource for that container?
I assume that for production we need to work with [`dpladm/bin/dpladm-shared.source` similar to this approach](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/main/infrastructure/dpladm/bin/dpladm-shared.source#L122-L123), but what options are there in the docker-compose file in the dpl-cms repo?

ITViking:
arent these nginx confs per environment: https://github.com/danishpubliclibraries/env-canary/tree/main/lagoon/conf/nginx ?

ath88:
> > How is the node application deployed?
> 
> By adding a service block to docker-compose.yml: https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/develop/docker-compose.yml#L144-L148
> 
> So this is running in the PR env's as it is now. Has be a part of the site template as well when it needs to hit production.

Seems like this goes back to being an issue with lagoon. We could ask them if they have a suggestion on how to add an ingress annotation.

ITViking:
closing this, as we have solved it for now

PR #531, opened 2024-11-27
Add vertical pod autoscalers

ITViking:
#### What does this PR do?
it adds vertical pod autoscalers to our existing autoscaler collection. 
The VPA's are installed in **Initial** mode. This ensures that the VPA does not collide with the existing HPA's. 
Instead of recreating a pod with the new recommended resources requests, it instead applies the new recoomandations when a pod is for some reason restartet. This would typically, in our case, be when we deploy. 
We should thus now not need to run the `adjust-resource-request` task anymore. 
Additionally we should now have more precise requests set for each environment and for each site. 

All in all I hope to help us have a more smooth deployment as well as a generally lower resource need. 

No. Just read it. 


infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: cli-vpa
+  namespace: ${SITE_NAME}-${SITE_ENV}
+spec:
+  targetRef:
+    apiVersion: "apps/v1"
+    kind: Deployment
+    name: cli
+  updatePolicy:
+    updateMode: "Initial"
```
ath88, line 12: I'd like to initially just deploy it with `Off` and then inspect the recommendations manually before having them applied.

infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
```diff
@@ -0,0 +1,21 @@
...
+  resourcePolicy:
+    containerPolicies:
+    - containerName: "cli"
+      minAllowed:
+        cpu: "4m"
+        memory: "20Mi"
+      maxAllowed:
+        cpu: "150m"
+        memory: "250Mi"
```
ath88, line 21: Nitpicking here - none of these doublequotes are necessary. They are clutter. :)

infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
ITViking: I did that in all the test projects on both their environments, and it looked robust and with good measure to me. 
I also tested on a couple of live sites. Do you still want them of for now?

infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
ITViking: ah... thanks... will fix

infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
ath88: Yes, lets have the observations documented first. :) I'd hate for this to blow up in our face and not having had four eyes on the recommendations first.

ITViking: :+1:

ath88 (approved):
There are still a few unnecessary double quotes, but that's still nitpicking.

Looking forward to seeing how it performs. :)

ITViking:
oh yeah... sorry bout that - will fix

ITViking:
done - removed the double qoutes

PR #534, opened 2024-11-28
Moduletestsyncing

xendk:
#### What does this PR do?

Adds `sites:webmaster:reset-moduletest` task and updates resetting moduletest runbooks.

We're getting requests to reset sigle sites, so a command that does exactly that is handy and eliminates the risk of fumbling the four required parameters to `site:environment:sync`.

Added note to resetting module rulebook that it's in fact not quite a proper reset.

Added the missing rulebook about using resetting moduletest sites using `dplsh` to aid discoverability.

### DIFF
--- a/docs/runbooks/reset-moduletest.md
+++ b/docs/runbooks/reset-moduletest.md
@@ -2,23 +2,21 @@
 
 ## When to use
 
-When a moduletest site need reseting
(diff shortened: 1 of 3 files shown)


ITViking (approved):
Terrific addition - let's go!

PR #541, opened 2024-12-04
Add page describing Lagoon projects

xendk:
#### What does this PR do?

Document the difference between Lagoon projects.


ITViking (approved):
good stuff. Maybe it is also time to have the runbooks chapterized properly, so information is logically divided

PR #543, opened 2024-12-05
Remove unused step

xendk:
#### What does this PR do?
It remove steps that we do not want to use at this moment, as they move the pod around. It is not possible for us to move pod around without interupting the live sites, as no PodDisruptionBudget has been installed


kasperg (commented):
Please consider adding a sentence or two on why this change is needed.

It might be a result of a conversation between you and @ITViking but it is good to have the summary of it for future reference included with the change.

It might come handy if someone wonders about the actual tasks which still remain within the project.

ITViking (approved):
looks good - let's go

xendk:
@ITViking You have a sentence or two on why we're removing them?

PR #544, opened 2024-12-05
Update editors ond webmasters to 2024.49.0

xendk:
Update editor and moduletest sites to 2024.49.0.

Also remove the hackery needed in the last deployment to get 2024.48.1 out on webmaster-with-defaults sites.


ITViking (approved):
looks good to me

PR #546, opened 2024-12-06
Turn off echo'ing for some tasks

xendk:
Tasks that contain a lot of shell code and echo their own status should be silenced to limit the amount of (useless) output.

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -110,6 +110,7 @@ tasks:
     env_repos:provision:
       deps: [_req_env, _env_repos:terraform:init]
       desc: Provision infrastructure for an environment
+      silent: true
       summary: |
         Use Terraform to provision the repositories for our environments.
       env:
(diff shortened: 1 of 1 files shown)


ITViking:
are we loosing any vital information here, which, if left in place, would allow us to now in which step something went wrong?

xendk:
@ITViking Nope, it just suppresses printing the command before running it. Basically avoiding filling your terminal with

```
          if [ -z "{{.SKIP}}" ]; then
            if [ "$(lagoon get project --project "{{.SITE}}" --output-json | jq '.data[0].id' --raw-output)" = "0" ]; then
              PROJECT_NAME="{{.SITE}}" GIT_URL="{{.GIT_URL}}" BRANCHES="{{.BRANCHES}}" task lagoon:project:add;
            else
              PROJECT_NAME="{{.SITE}}" GIT_URL="{{.GIT_URL}}" BRANCHES="{{.BRANCHES}}" task lagoon:project:update;
            fi
          else
            echo "Skipped project ensure."
          fi
```

Any output from the command is still output.

ITViking:
I would like one more take on this  @kasperg

ITViking (approved):
let's try it out

PR #547, opened 2024-12-06
Add bnf

xendk:
Add BNF site, and update documentation a bit.

### DIFF
--- a/infrastructure/environments/dplplat01/sites.yaml
+++ b/infrastructure/environments/dplplat01/sites.yaml
@@ -58,6 +58,12 @@ sites:
     plan: webmaster
     deploy_key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHvhy79hHjLcQJCcMNwci1Q/P/O2LwD4IzBVfkmRGKom
     <<: *webmaster-release-image-source
+  # BNF site.
+  bnf:
+    name: "Bibliotekernes Nationale Formidling"
+    description: "The BNF content sharing site"
+    <<: *default-release-image-source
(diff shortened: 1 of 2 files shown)


infrastructure/environments/dplplat01/sites.yaml
```diff
@@ -58,6 +58,14 @@ sites:
     plan: webmaster
     deploy_key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHvhy79hHjLcQJCcMNwci1Q/P/O2LwD4IzBVfkmRGKom
     <<: *webmaster-release-image-source
+  # BNF site.
+  bnf:
+    name: "Bibliotekernes Nationale Formidling"
+    description: "The BNF content sharing site"
+    primary-domain: bnf.dplplat01.dpl.reload.dk
```
hypesystem, line 65: If the only domain used is the auto-generated domain, you shouldn't explicitly set a primary domain.

See example of going from autogenerated domains to live/prod domains here: https://github.com/danskernesdigitalebibliotek/dpl-platform/commit/3f3a7ea8830bc430dbbb0e024943312d35108543

hypesystem (changes requested):
Looks good :smile: nice updates to the docs.

A small note on the bnf site setup: you shouldn't set `primary-domain` if you are only using auto-generated routes.

infrastructure/environments/dplplat01/sites.yaml
xendk: Dammit, I actually thought about removing it, but forgot to do it.

PR #553, opened 2024-12-09
Wait a bit between namespaces during resource adjustments

kasperg:
#### What does this PR do?

Add 15 seconds of wait between resource requests. 

With ~100 sites this will add a total of ~25 minute delay for the deployment process. It is hard to tell whether that is enough.

It wold be optimal if we could either 

1. Warm the caches site by site manually - e.g. by running `drush cr` or curl'ing `/` or `health`
2. Avoid flushing the cache during resource request adjustments e.g. by having a persistent cache or by moving the cache away form each site


ITViking:
This is now redundant as we have installed Vertical Pod Autoscalers in the cluster

PR #560, opened 2024-12-15
automatically adjust pr envs resource request

ITViking:
#### What does this PR do?
This adds a cronjob, which will run every 6 hours. It will take approximately pr-envs*1'ish minute to complete. 
The job will adjust the redis, varnish, nginx and cli deployments to create pods with a set resource request, fitting for pr environments. 

This is in effect now


infrastructure/environments/dplplat01/configuration/vertical-pod-autoscaler/patch-cms-bnf-deployments.yaml
```diff
@@ -0,0 +1,88 @@
...
+                        memory_limit="125Mi"
+                        ;;
+                      *)
+                        memory_request="150Mi"
+                        memory_limit="256Mi"
+                        ;;
+                    esac
+                    kubectl patch deployment $deploy -n $ns \
+                      --type=json \
+                      -p="[{\"op\": \"replace\", \"path\": \"/spec/template/spec/containers/0/resources\", \"value\": {\"requests\": {\"cpu\": \"15m\", \"memory\": \"$memory_request\"}, \"limits\": {\"cpu\": \"200m\", \"memory\": \"$memory_limit\"}}}]"
```
ath88, line 48: Does the output here reveal if the deployment was changed or not? If so - you could skip the wait, if the resources had already been set.

ath88:
I like your approach. Could you turn it into a helm chart instead, so the relationship between the resources is clearly marked for someone inspecting the cluster?

infrastructure/environments/dplplat01/configuration/vertical-pod-autoscaler/patch-cms-bnf-deployments.yaml
```diff
@@ -0,0 +1,88 @@
...
+kind: ClusterRole
+metadata:
+  name: namespace-lister
+rules:
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["list"]
+- apiGroups: ["apps"]
+  resources: ["deployments"]
+  verbs: ["get", "list", "patch"]
```
ath88, line 75: This does more than just listing namespaces. Perhaps name the resource for what uses them instead. This will be easier if you convert it to a helm chart.

infrastructure/environments/dplplat01/configuration/vertical-pod-autoscaler/patch-cms-bnf-deployments.yaml
ITViking: that is a good point, will have to look into that

infrastructure/environments/dplplat01/configuration/vertical-pod-autoscaler/patch-cms-bnf-deployments.yaml
ITViking: yeah... sorry about that. I started out with a logging functionality to test it out and get a quick feedback loop. Will fix

ITViking: fixed

infrastructure/environments/dplplat01/configuration/vertical-pod-autoscaler/patch-cms-bnf-deployments.yaml
ITViking: I left a comment, but there's more pressing matters to get at than this at the moment

infrastructure/environments/dplplat01/configuration/patch-pr-env-resource-requests/chart/templates/patch-cms-bnf-deployments.yaml
```diff
@@ -0,0 +1,89 @@
...
+  namespace: default
+spec:
+  schedule: "0 */6 * * *" # Every 6 hours
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: patch-resources
+            image: bitnami/kubectl:latest
```
hypesystem, line 14: Pin to kubernetes version we know works with our setup.

kubectl may update or we may update cluster to create inconsistency unknowingly - so it's better to be able to control it.

infrastructure/environments/dplplat01/configuration/patch-pr-env-resource-requests/chart/templates/patch-cms-bnf-deployments.yaml
```diff
@@ -0,0 +1,89 @@
...
+                    # TODO: we could look into away of checking if the patching of a deployment went well, so we dont have to sleep
+                    kubectl patch deployment $deploy -n $ns \
+                      --type=json \
+                      -p="[{\"op\": \"replace\", \"path\": \"/spec/template/spec/containers/0/resources\", \"value\": {\"requests\": {\"cpu\": \"15m\", \"memory\": \"$memory_request\"}, \"limits\": {\"cpu\": \"200m\", \"memory\": \"$memory_limit\"}}}]"
+                  else
+                    echo "Deployment $deploy not found in namespace $ns"
+                  fi
+                done
+                echo "sleeping for a minute to give the deployments time to get back up and not crash the database"
+                sleep 60
```
hypesystem, line 55: Hmm, I think my take on this is that it is potentially creating a lot of work on the Kubernetes API server if we patch everything every minute?

Could we find a way to directly query if the deployments have their memory limits set correctly before trying to set them?

Looking into this, will report back if I find a good approach.

hypesystem (approved):
Looks very reasonable to me! Have you tested that it works in some limited capacity? Or are you awaiting approval before testing?

I have one concern which I will see if I can help with a solution for.

infrastructure/environments/dplplat01/configuration/patch-pr-env-resource-requests/chart/templates/patch-cms-bnf-deployments.yaml
hypesystem: Beginning of something:

```sh
kubectl get deploy -n $ns -o json | jq '.items[] | { name: .metadata.name, resources: .spec.template.spec.containers[0].resources }'
```

produces a list of JSON objects each with a name and the resource spec. Next up, picking those that do not match the desired memory limit and request.

hypesystem:
```sh
kubectl get deploy -n $ns -o json | jq '.items[] | [.metadata.name, .spec.template.spec.containers[0].resources.limits.memory, .spec.template.spec.containers[0].resources.requests.memory] | join(",")' --raw-output
```

This outputs a list of the format:

```sh
<deploymentname>,<memorylimit>,<memoryrequest>
```

To get a particular list of deployments, we can append this to the command (selecting `nginx|redis|varnish|cli` followed by memory indications):

```sh
 | grep -E '^(nginx|redis|varnish|cli),([0-9]+[A-Za-z]+|),([0-9]+[A-Za-z]+|)$'
```

If we save the output of the above as variable `spec` get the individual fields like this:

```sh
deployment_name=`echo $spec | cut -f1 -d,`
current_memory_limit=`echo $spec | cut -f2 -d,`
current_memory_request=`echo $spec | cut -f3 -d,`
```

Now we can decide to only patch if the limit or request differ from the current state. Much fewer requests.

When everything is as it should be it means we get the following requests every execution (6 hours):
- 1 GET namespaces
- 1 GET deployments per namespace
- (and then up to 4 patch calls per namespace, but for most sites 0)

Whereas before, when we naively patched, we would get:
- 1 GET namespaces
- 4 PATCH calls per namespace

ITViking:
This has been tested and has been live since before christmas (december? can't remeber, but for a good while now)

hypesystem:
@ITViking Then I think this is good to merge - we can look at potentially using my research at a later point if we think it is too resource intensive. The minute long break between each namespace is probably helping us though :smile:

ITViking:
sounds good :)

PR #570, opened 2025-01-08
Stabilize moduletest environment synchronization for webmaster sites

kasperg:
#### What does this PR do?

Stabilize the synchronization process from main to module test environments for webmaster sites.

Practice shows there have been a number of problems with the synchronization process. This primarily seems to be caused by idle timeouts in the SSH connection to Lagoon combined with large amount of files which needs to be transferred causing the process to stop. This is combined with the error handling which in the current setup just means that we continue with the next site.

To address this this PR does three things:

1. Split the process into individual task/`lagoon ssh` commands to limit the scope
2. Exclude generated files form the synchronization process to reduce the amount of data that needs to be synchronized
3. Change error handling to abort execution if commands fail.

Try running the process for a webmaster site e.g. `PROJECT=customizable-canary task sites:webmaster:reset-moduletest`

I recommend going through the PR commit by commit.

### DIFF
--- a/infrastructure/Taskfile.yml
+++ b/infrastructure/Taskfile.yml
@@ -1136,7 +1136,7 @@ tasks:
         TARGET_PROJECT: "{{.TARGET_PROJECT}}"
         TARGET_ENV: "{{.TARGET_ENV}}"
       cmds:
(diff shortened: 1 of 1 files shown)


xendk (approved):
:+1: 

Maybe a `&&` instead of `;` between drop and sync.

ITViking:
@kasperg I'm for some reason not able to run this now - could you give it a go and let me know if it works for you?

ITViking:
I can't seem to make this work correctly at the moment. I would like one of you guys to  update your `dplsh`to the latest version (0.11) and run it again and tell me if it works. Maybe it's an issue with my machine

PR #572, opened 2025-01-09
display dplsh version on start up

ITViking:
#### What does this PR do?
Shows dplsh's version on start up. This will make it easier for us to know which version we are running and enable troubleshooting
image

Please do, simply restart the shell


xendk (approved):
:+1:

PR #577, opened 2025-01-14
Deploy release 2025.3.0 on staging

kasperg:
#### What does this PR do?

Deploy release 2025.3.0 on staging.

This has already been applied.

Note that this PR also includes a few other changes that are needed before the update can be deployed. This includes:

- Adding a new site for Solrød because a table for the current database has been corrupted
- Removing a call to `lagoon whoami` which currently fails


ITViking (approved):
LGTM

PR #616, opened 2025-03-06
Release 2025.10.0

xendk:
Deploy release 2025.10.0 to editor libraries and moduletest envs, 2025.9.0 to the main env on webmaster libraries.

And fix quoting in runbook (which is out of date, but that's another thing).


spaceo (approved):
LGTM 👍

PR #638, opened 2025-03-28
allow cli to use more memory and provision some more for it from the …

ITViking:
…get go

#### What does this PR do?
It allows all production sites CLI pod to use more memory

Just read it - it will be applied on the next deployment when merged 

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-324

### DIFF
--- a/infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
+++ b/infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
@@ -14,8 +14,8 @@ spec:
     containerPolicies:
     - containerName: cli
       minAllowed:
-        cpu: 4m
-        memory: 2Mi
+        cpu: 200m
+        memory: 200Mi
       maxAllowed:
-        cpu: 150m
-        memory: 250Mi
+        cpu: 400m
+        memory: 1.1Gi
--- a/infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
+++ b/infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
@@ -14,8 +14,8 @@ spec:
     containerPolicies:
     - containerName: cli
       minAllowed:
-        cpu: 200m
-        memory: 200Mi
+        cpu: 4m
+        memory: 2Mi
       maxAllowed:
         cpu: 400m
         memory: 1.1Gi
--- a/infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
+++ b/infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
@@ -14,8 +14,8 @@ spec:
     containerPolicies:
     - containerName: cli
       minAllowed:
-        cpu: 4m
-        memory: 2Mi
+        cpu: 20m
+        memory: 300Mi
       maxAllowed:
         cpu: 400m
         memory: 1.1Gi


infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
```diff
@@ -14,8 +14,8 @@ spec:
     containerPolicies:
     - containerName: cli
       minAllowed:
-        cpu: 4m
-        memory: 2Mi
+        cpu: 200m
+        memory: 200Mi
```
kasperg, line 18: Why do we need to raise the minimum?

We we discussed increased resource consumption I expect that raising the minimum resource allocation will have a much larger impact than raising the maximum allowed usage.

If this is related to vertical pod autoscaling then I would suggest that scaling is turned off for the CLI. If the VPA sets the maximum too low then it will effectively kill mid-run just as we are seeing now. As I see it it would be preferable that the memory limits are fixed.

infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
```diff
@@ -14,8 +14,8 @@ spec:
...
       maxAllowed:
-        cpu: 150m
-        memory: 250Mi
+        cpu: 400m
+        memory: 1.1Gi
```
kasperg, line 21: Is there a reason why this is raised to 1.1Gi when we requested 1Gi?

It is not the end of the world. I am just wondering :smile:

kasperg (commented):
> It allows all production sites CLI pod to use more memory

Does this change really only target production CLI pods? I cannot infer this from the code change.

As I see it we need to apply this change for both production and non-production CLI pods. Otherwise we cannot test background jobs properly before they reach production as they may be killed mid-run due to excessive memory usage.

ITViking:
It only targets production environments. The reason being the way it was implemented, matching the horizontal pod autoscaler. The VPA's are set when we deploy. 

Therefore this is only on half of the task, the other half of course being the pr envs. Remember - we're doing it this way because Lagoon bars us from doing it the right way.

infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
ITViking: We did discuss just raising the limit and see if that is enough

xendk: After reading up on VPA, I have a question: What's stopping the VPA from killing the pod to adjust limits mid-cron-run?

As I read it, the *minimum* ought to be raised to at least 400MiB, else we risk the VPA scaling the container down to below 400 due to inactivity, which then triggers k8n to kill the container before `drush` reaches its limit when encountering load.

infrastructure/dpladm/vpa-templates/cli-vpa.template.yaml
xendk: > recommand killing the pod

Recommend? Then who takes de decision?

ITViking:
The the updater (thr object that would kill pod surpassing their limits to enforce them) is not enabled. Thus Recommendations are implemented when pod die for any reason. Ie. Release deployments or errors killing the pod

Med venlig hilsen/Best regards

Philip Dein
Copenhagen, DK
+45 2840 8442

Sent from Proton Mail Android

-------- Original Message --------

xendk:
So, if the updater doesn't run...

A new env is created, and the VPA doesn't know anything yet, so it gets the default 125MiB... So cron jobs might initially fail, having k8n kill them, and then, hopfully, the VPA will have noticed and upscaled them.

But as I understand [the docs](https://learn.microsoft.com/en-us/azure/aks/vertical-pod-autoscaler-api-reference#containerresourcepolicy) the VPA adjusts requests, but it's the limit that's killing us?

ITViking:
The VPA has 3 components. We do not have the `Updater` enabled. (https://www.kubecost.com/kubernetes-autoscaling/kubernetes-vpa/#components-of-vpa-6) 

If I understand the question correctly, you're asking what will happen when a new Environment (new site) is added. You are correct, all the services will not at that point be adjusted correctly, since the VPA Recommender hasen't had a chance to monitor the pod running yet. They will have the resource request and limits set when the pods are killed next time. This gives a single handle to pull: Kill a pod we want to run with the adjusted resource allowance. 
This will be an issue for you devs of the PR envs, and something I am going to have to figure out a solution for.

xendk:
But doesn't the VPA set the *requested* resources, and not the *limit*?

ITViking:
It sets both

xendk:
> This gives a single handle to pull: Kill a pod we want to run with the adjusted resource allowance.

As far as I can tell, timing is rather important here. If we kill the pod before cron has run, the VPA haven't seen much usage, so won't give it much. Unless we assure that `minAllowed` is "enough".

> It sets both if you ask it to, which we have

The only explanation I've been able to find says it [sets it proportionally to the existing configuration](https://medium.com/infrastructure-adventures/vertical-pod-autoscaler-deep-dive-limitations-and-real-world-examples-9195f8422724).

The request and limits starts out at 25/125MiB.

So with the proposed changes, if the VPA haven't seen exceptional usage yet, it'll set the request to the minimum, 200MiB, and scale the limit (200/25*125) to 1000MiB.

Which basically makes the bumping of `maxAllowed` irrelevant, as 1000MiB is way above the 400MiB limit that PHP enforces.

So the minimal change needed to avoid `drush cron` being killed ought to be just raising the `minAllowed` memory to 100MiB, as that gives a limit of 500MiB, which ought to be enough, for `drush cron` at least. Changing `maxAllowed` shouldn't matter. This doesn't account for the overhead of cron process, etc, but that's difficult to gauge. `free` in a running cli container claims its using 12GiB, so k8n is basing its calculations on something else.

ITViking:
@xendk I agree with you, let's give it some more out the gate

xendk:
So, raise the minimum to 200 and leave the maximum as is?

That leaves the question of how do we get the VPA to set the bigger limit and k8n to restart the service automatically and preferably before anyone attempts to use the pod?

ITViking:
I thinking 300Mi. We should enable the updater and for the CLI change the `updateMode` to auto, so it is instantly caught when we release 
I'm still working on how get the development envs targeted by this

xendk:
> I'm thinking 300Mi

:+1: 

> We should enable the updater and for the CLI change the updateMode to auto, so it is instantly caught when we release

As I read the documentation, it'll still wait some time while it collects stats. It might just be 5 minutes, but it's still an annoying gap.

ITViking:
@xendk agreed - it would be nice if Lagoon simply let us do it, how we wanted

ITViking:
no longer relevant as it seems the VPA i recommending way to much memory

PR #645, opened 2025-04-03
If project uses go set the subdomains and allow bigger cookies for node service

ITViking:
#### What does this PR do?
The GO project is going to be available on a `go.` subdomain. We need to calculate this subdomain and allow a bigger cookie than usual to be passed through to the underlying nodejs service. 
We do not want to calculate the subdomain for libraries who don't use the GO project. 

This implementation will figure out how to place the `.go` subdomain correctly on both non-www domains as well as www domains. 
It will also redirect from secondary GO subdomains to the primary one, just as we and the libraries are used to. 

This should be tested and can be tested in the following way: 

add the following to the canary site's `sites.yaml` config: 
   ```
 primary-domain: www.bibliotest.deranged.dk
 secondary-domains:
    - bibtest.deranged.dk
    - www.bibtest.deranged.dk

```

Then run `SITE=canary task site:sync` in DPLSH. 
After that check this link just beneath the line: https://github.com/danishpubliclibraries/env-canary/blob/main/.lagoon.yml#L72

There shouldn't be anything. 

Now add the following property to canary in `sites.yaml`: 
```
hasGo: "test"
``` 

There still shouldn't be anything. 

Change `hasGo`'s value to ` "false"` 

Still nothing should show. 
Now change it to `"true"` or `true` and now routes and annnotations should show up. 

Read and test it

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?label=Brahma&selectedIssue=DDFDRIFT-323

### CODE REVIEW

File: infrastructure/dpladm/bin/dpladm-shared.source [hunk 1]
```diff
@@ -111,6 +115,7 @@ EndOfMessage
     PRIMARY_DOMAIN="${singleRouteIndent}- \"${primaryDomain}\""
   fi
 
+
```
ath88, line 118: Why this?

infrastructure/dpladm/bin/dpladm-shared.source
```diff
@@ -76,6 +76,10 @@ function renderProfileTemplate {
   local primaryDomain=${8:-}
   # A space-seperated list of secondary domains (optional)
   local secondaryDomainsString=${9:-}
+  # The primary GO subdomain (optional)
+  local primaryGoSubDomain=${10:-}
+  # A space seperated list of secondary GO sub domains (optional)
+  local secondaryGoSubDomains=${11:-}
```
ath88, line 82: What happens if there are no secondary domain strings?

infrastructure/dpladm/bin/dpladm-shared.source
```diff
@@ -138,13 +170,15 @@ EndOfMessage
...
+  export PRIMARY_GO_SUBDOMAIN_AND_ANNOTATION
+  export SECONDARY_GO_SUBDOMAINS
   export IMPORT_TRANSLATIONS_CRON=${importTranslationsCron}
 
   # TODO this templating could be quite a bit more flexible.
   # Tell envsubst which variables to replace. This allows other variables to
   # remain untouched.
   # shellcheck disable=SC2016
-  local variablesToSubst='$RELEASE_IMAGE_REPOSITORY $RELEASE_IMAGE_NAME $RELEASE_TAG $ENABLE_ROUTES $LAGOON_IMAGES_RELEASE_TAG $LAGOON_PROJECT_NAME $LAGOON_ENVIRONMENT $PRIMARY_DOMAIN $SECONDARY_DOMAINS $IMPORT_TRANSLATIONS_CRON'
+  local variablesToSubst='$RELEASE_IMAGE_REPOSITORY $RELEASE_IMAGE_NAME $RELEASE_TAG $ENABLE_ROUTES $LAGOON_IMAGES_RELEASE_TAG $LAGOON_PROJECT_NAME $LAGOON_ENVIRONMENT $PRIMARY_DOMAIN $SECONDARY_DOMAINS $IMPORT_TRANSLATIONS_CRON $PRIMARY_GO_SUBDOMAIN_AND_ANNOTATION $SECONDARY_GO_SUBDOMAINS'
```
ath88, line 181: It looks like `$IMPORT_TRANSLATIONS_CRON` is after the secondary domains variable - is there an off-by-one error here with variable parsing?

infrastructure/dpladm/bin/sync-site.sh
```diff
@@ -97,6 +97,70 @@ function getSiteSecondaryDomains {
...
+        echo "true"
+        return
+    fi
+    if [[ "${hasGo}" == "false" ]]; then
+        echo "false"
+        return
+    fi
+    echo "false"
+    return
+}
```
ath88, line 117: I would recommend only going with `true` as a true value - having several positive values only makes things more confusing when people see values being used in the yaml-file.

ath88: Also - generally when making values-files - its only the boolean value that is used as a true value. The few times you use the string `"true"` is when the value is passed to another system directly, like eg. nginx annotations.

infrastructure/dpladm/bin/sync-site.sh
```diff
@@ -97,6 +97,70 @@ function getSiteSecondaryDomains {
...
+function calcutelateSecondaryGoSubDomains {
+    local hasGo=$(projectHasGo "${1}" "${2}")
+    if [[ "${hasGo}" = "false" ]]; then
+        echo ""
+        return
+    fi
+
+    local secondaryDomains=$(getSiteSecondaryDomains "${1}" "${2}")
+    if [[ "${secondaryDomains}" == "null" ]]; then
+        echo "shit"
```
ath88, line 145: uhm?

infrastructure/dpladm/bin/dpladm-shared.source
ath88: I see you'll then just have empty double quotes - right?

ath88 (changes requested):

infrastructure/dpladm/bin/dpladm-shared.source
ITViking: yes

infrastructure/dpladm/bin/sync-site.sh
ath88: I see your change. Looks good.

infrastructure/dpladm/bin/dpladm-shared.source
ITViking: to create some structure in the file, making it clear, that this is something new from the line above

ath88 (approved):

PR #659, opened 2025-04-10
set min and maxAllowed for other containers that run in the CLI pod

ITViking:
#### What does this PR do?
I noticed that the CLI pod accross namespaces, were allowed a lot of resouces, somtimes reaching an allowance of over a terrabyte of memory, but in most cases just being allowed to use up to 33Gb of memory. This PR seeks to have the CLI pod's VPA recommend some more sane defaults.

Just read it

Just read it 


ath88 (approved):
Looks fine - CLI pods is such a bad idea anyway, but this at least makes it better.

PR #690, opened 2025-05-20
we need to set primary domains for these sites inorder for adgangstje…

ITViking:
…nesten and uniLogin to whitelist the domains so we have fully fledged GO apps on each project

#### What does this PR do?
It sets primary domains for:
- canary
- cms-school
- staging

we need these so we can have a fully fledged GO app for each of these projects. Otherwise we can't sign in on the each projects GO app

Just read it

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-358


Dresse (approved):
LGTM 👍

PR #693, opened 2025-05-22
Revert "we need to set primary domains for these sites inorder for adgangstje…"

kasperg:
Reverts danskernesdigitalebibliotek/dpl-platform#690

These changes update which urls the sites are accessible at. They have not been whitelisted at Adgangsplatformen yet and this makes it impossible to login as a patron until this has happended. This in turn prevents testing.


ITViking (approved):
let's go

PR #696, opened 2025-05-23
Corrections to getting go live

ITViking:
#### What does this PR do?
This PR makes a few corrections to have we add a GO project to a site

Just read it - can be tested on canary

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-358

### DIFF
--- a/infrastructure/environments/dplplat01/sites.yaml
+++ b/infrastructure/environments/dplplat01/sites.yaml
@@ -35,7 +35,6 @@ sites:
     releaseImageName: dpl-cms-source
     dpl-cms-release: "2025.21.2"
     plan: webmaster
-    primary-domain: canary.dplplat01.dpl.reload.dk
     # TODO: Remove this once the the Adgangsplatformen OpenID client used here
     # has been updated to use the primary domain.
     autogenerateRoutes: true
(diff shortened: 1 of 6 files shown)


infrastructure/environments/dplplat01/sites.yaml
```diff
@@ -47,6 +46,9 @@ sites:
     releaseImageRepository: ghcr.io/danskernesdigitalebibliotek
```
Dresse, line 46: This part is done to bypass some roadblocks while trying to setup a working GO environment. This should be changed back at some point.

Dresse (approved):
LGTM. Have 1 comment that we need to consider. 👍

infrastructure/environments/dplplat01/sites.yaml
ITViking: I'm not sure I quite follow - can you elaborate?

Dresse: I can see I somehow commented the wrong line. 

I was just commenting on that we probably need to change the primary-domain back, so `canary` have `canary.dplplat01.dpl.reload.dk` as the `primary-domain`.

Hope it clears it up.

PR #710, opened 2025-06-02
Deploy cms ver. 2025.22.0 and Go 0.25.13 for go-pilots/bnf

spaceo:
#### What does this PR do?
Deploys cms ver. 2025.22.0 and Go 0.25.13 for go-pilots/bnf
Created defaults for go pilots and the BNF site so it is easier to maintain.


ITViking (approved):
Good to go

PR #714, opened 2025-06-03
Set env variables on project level

ITViking:
#### What does this PR do?
We want to clean up the environment variables, so we don't have duplicates and  only use them on the project level, when that is what makes sense 

This can't easily be tested.

Just read it 

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-366


spaceo (approved):
LGTM 👍

PR #716, opened 2025-06-04
Automate environment level variables removal

ITViking:
#### What does this PR do?
This script removes the environment level variables specified in the list from all the sites, main and moduletest.

Just read it 

https://reload.atlassian.net/jira/software/c/projects/DDFDRIFT/boards/464?selectedIssue=DDFDRIFT-366


infrastructure/task/scripts/delete-environment-level-variables.mjs
```diff
@@ -0,0 +1,41 @@
...
+    } catch (error) {
+      throw Error("failed to delete variables for BNF and GO secrets", { cause: error });
+    }
+  }
+}
+
+async function isWebmaster(project) {
+   const result = await $`cat ../host_mount/environments/dplplat01/sites.yaml | yq '.sites.${project}.plan'`;
+   return result.stdout === "webmaster\n" ? true : false;
+}
```
spaceo, line 30: Would it be an idea to have common functions like this in some kind of utility module and import them when needed.

PR #748, opened 2025-06-24
Added .idea files to .gitignore

Dresse:
#### What does this PR do?
Added `.idea` files generated by PHPStorm to `.gitignore`, as we do not want them in Git, and it is annoying to skip them on git diffs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 **/kdevconfig.yaml
 **/skaffold.yaml
 .env
+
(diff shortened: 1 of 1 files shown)


xendk (changes requested):
[#dev snak > Global gitignore @ 💬](https://reload.zulipchat.com/#narrow/channel/235451-dev-snak/topic/Global.20gitignore/near/525455336)

PR #761, opened 2025-07-02
Update sites sync and env var setter task

ITViking:
Running `task sites:sync` as it currently is, runs a lot of subtasks, which we don't usually want or need. 
This changes `sites:sync` so that it doesn't. Instead the user now has to actively decide to run the subtasks of `sites:sync`. 

Additionally this corrects and cleans up some environments variables, that we are using. 
We still have a task to solve here, which is to store theses secrets somewhere, where we can update them from


xendk:
I don't like it. The idea of the task is to be idempotent so it's safe to run if you haven't touched anything that triggers a change in a particular subtask. If you make it explicit, we're making it a requirement for everyone that might ever deploy to know exactly what each subtask does and when to apply it. Which in turns means that everybody needs to know what a deploy key is and when it should be captured. You're raising the bar for when someone can be trusted to run the task significantly.

ITViking:
So what do you suggest we do instead

infrastructure/Taskfile.yml
```diff
@@ -1075,10 +1075,10 @@ tasks:
...
           vars:
             SITE: "{{.ITEM}}"
-            SKIP_ENSURE_PROJECT: "{{.SKIP_ENSURE_PROJECT}}"
-            SKIP_CAPTURE_DEPLOY_KEY: "{{.SKIP_CAPTURE_DEPLOY_KEY}}"
-            SKIP_PROVISION: "{{.SKIP_PROVISION}}"
-            SKIP_FIRST_DEPLOYMENT: "{{.SKIP_FIRST_DEPLOYMENT}}"
+            SKIP_ENSURE_PROJECT: "{{.SKIP_ENSURE_PROJECT | default TRUE}}"
+            SKIP_CAPTURE_DEPLOY_KEY: "{{.SKIP_CAPTURE_DEPLOY_KEY | default TRUE}}"
+            SKIP_PROVISION: "{{.SKIP_PROVISION | default TRUE}}"
+            SKIP_FIRST_DEPLOYMENT: "{{.SKIP_FIRST_DEPLOYMENT | default TRUE}}"
```
kasperg, line 1081: > Running task sites:sync as it currently is, runs a lot of subtasks, which we don't usually want or need.

I think this is a mistake.

I do not mind running subtasks even if they are usually no-ops compared to adding complexity to knowing when you should run them.

The problems we are seeing with environment variables is because we have introduced a new concept for setting these across applications without considering other workflows.

For example we cannot set `NEXT_PUBLIC_GO_GRAPHQL_CONSUMER_USER_PASSWORD` for a new site to the same value as is used on the other sites as [the value is generated at task runtime](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/da4550bd16dc1fe2b2cc8cca0cf0e9465806c9a2/infrastructure/task/scripts/create-go-and-bnf-consumer-and-secrets.mjs#L19-L19).

PR #818, opened 2025-08-05
set diskSize for all sites. Kobenhavn gets 15GB everyone else gets 10GB

ITViking:
Set diskSize for every single library. Kobenhavn gets 15 GB and everyone else gets 10GB

### DIFF
--- a/infrastructure/environments/dplplat01/sites.yaml
+++ b/infrastructure/environments/dplplat01/sites.yaml
@@ -43,6 +43,7 @@ sites:
     releaseImageRepository: ghcr.io/danskernesdigitalebibliotek
     releaseImageName: dpl-cms-source
     plan: webmaster
+    diskSize: 5
     dpl-cms-release: 2025.27.2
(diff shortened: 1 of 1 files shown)


spaceo (changes requested):
Two things:

1:
Although this is the right way to go, I think the repeated values calls for some defaults.
We could create a `disksize-[small|medium|large]` handle and use it like:
```
aero:
    name: "Ærø Folkebibliotek"
    description: "The library site for Ærø"
    deploy_key: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIDJNI7OU7O3X0AYNJSyZ9VkdSGaD5mAQmhcRMW72TpKi"
    primary-domain: www.arrebib.dk
    diskSize: 10
    secondary-domains:
      - arrebib.dk
    autogenerateRoutes: true
    << : [ *default-release-image-source, *disksize-medium ]
```
2:
I think BNF will need space like Kbh does. So I think we should allocate the same amount of space for the two sites

PR #821, opened 2025-08-06
Fix correct argument/parameter sequence

ITViking:
This fixes moduletests thinking they should have a GO project. 
The reason: Mixed up sequence of arguments/parameters


spaceo (approved):
LGTM 👍

PR #853, opened 2025-08-21
remove unused solrod site entry from sites.yaml

ITViking:
After talking it over with @kasperg, we have decided to remove the old solrod site which had it's database corrupted during an attempted database recovery by Azure. 

We are removing for two reasons:
1) The site is unused and it might confuse new people to the platform
2) The site doesn't pass deployment succesfully and thus creates noise when deciding if a release is complete


kasperg (approved):
:+1: Do we also have to remove the GitHub repo and/or Lagoon project manually? 

Normally (some of) this would happen through Terraform but we do not run this through deployments at the moment.

ITViking:
We should run over you list and ensure that we remove the site from Terraform, Lagoon  and Github

PR #890, opened 2025-09-30
Deploy 2025.40.0 to staging

xendk:
Deploy 2025.40.0 to staging and bnf moduletest. It has been applied.

Remove `moduletest-php-version` from staging, as both main and moduletest now needs 8.3 and are covered by the `php-version`.

Remvoe the same from bnf as `x-defaults` defines the proper version for both branches now.


kasperbirch1 (approved):
LGTM 👍

PR #894, opened 2025-10-02
remove cert-manager-verifier

ITViking:
I suggest to remove cert-manager-verifier from DPLSH because we're not using it. Removing it will also make it easier to get out DPLSH down the road - and for good measure: less code is better


ath88 (approved):
Approved - but it would be good to have a motivation in the PR description as well.

PR #967, opened 2026-01-09
Deploy hotfix 2026.2.2

xendk:
Apply 2026.2.2 to production sites affected.

Also apply it to all moduletest sites. As moduletest sites aren't production sites, it's not worth the effort to special case the affected sites.


rasben (approved):

xendk:
Superceeded by another hotfix.

PR #1061, opened 2026-03-06
Node 24 + Make node version configurable.

rasben:
This requires a hotfix for the webmasters GO release.

https://reload.zulipchat.com/#narrow/channel/240325-DDF/topic/node.2022/with/572404876

### DIFF
--- a/infrastructure/dpladm/bin/dpladm-shared.source
+++ b/infrastructure/dpladm/bin/dpladm-shared.source
@@ -80,12 +80,14 @@ function renderProfileTemplate {
   local diskSize=${10:-}
   # The version of PHP to use
   local phpVersion=${11:-}
+  # The version of Node to use
+  local nodeVersion=${12:-}
   # The primary GO subdomain (optional)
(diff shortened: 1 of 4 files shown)


ITViking (approved):
let's go

ITViking (dismissed):
approved

kasperg (changes requested):
Making the node version configurable is fine to me but are we ready to deploy Node 24 with the next release?

https://github.com/danskernesdigitalebibliotek/dpl-web/pull/209 is still open. I would prefer that we merge the functionality here but keep the Node version steady. Then as we roll out the Node 24 upgrade we can bump the versions with the related change to the Go release.

rasben:
@kasperg Good point.
To avoid confusion, I've closed this PR, and instead opened these two:

https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/1084

https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/1085

PR #1084, opened 2026-03-26
Make Node version configurable, like PHP.

rasben:
We will soon need to upgrade node.
This will require a two-step phase, as webmaster sites will run on a prior version than other sites.

### DIFF
--- a/infrastructure/dpladm/bin/sync-site.sh
+++ b/infrastructure/dpladm/bin/sync-site.sh
@@ -277,5 +277,5 @@ set -o errexit
 syncEnvRepo "${SITE}" "${releaseTag}" "${BRANCH}" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionMain}" "${nodeVersion}" "${primaryGoSubDomain}" "${secondaryGoSubDomains}" "${goRelease}"
 
 if [ "${plan}" = "webmaster" ] && [ "${BRANCH}" = "main" ]; then
-    syncEnvRepo "${SITE}" "${wmReleaseTag}" "moduletest" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionModuletest}" "${nodeVersion}"
+    syncEnvRepo "${SITE}" "${wmReleaseTag}" "moduletest" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionModuletest}"
 fi
--- a/infrastructure/dpladm/bin/sync-site.sh
+++ b/infrastructure/dpladm/bin/sync-site.sh
@@ -157,21 +157,15 @@ function getPhpVersion {
 }
 
 function getNodeVersion {
-    local branchNodeVersion nodeVersion
+    local nodeVersion
 
-    branchNodeVersion=$(yq eval ".sites.${1}.${2}-node-version" "${3}")
-    if [[ -n "${branchNodeVersion}" && "${branchNodeVersion}" != "null" ]]; then
-        echo "${branchNodeVersion}"
-        return
-    fi
-
-    nodeVersion=$(yq eval ".sites.${1}.node-version" "${3}")
+    nodeVersion=$(yq eval ".sites.${1}.node-version" "${2}")
     if [[ -n "${nodeVersion}" && "${nodeVersion}" != "null" ]]; then
         echo "${nodeVersion}"
         return
     fi
 
-    echo "Error: Neither specific or general node-version is defined for environment ${2}"
+    echo "Error: node-version is not defined"
     exit 1
 }
 
@@ -275,15 +269,13 @@ phpVersionMain=$(getPhpVersion "${SITE}" "main" "${SITES_CONFIG}")
 failOnErr $? "${phpVersionMain}"
 phpVersionModuletest=$(getPhpVersion "${SITE}" "moduletest" "${SITES_CONFIG}")
 failOnErr $? "${phpVersionModuletest}"
-nodeVersionMain=$(getNodeVersion "${SITE}" "main" "${SITES_CONFIG}")
-failOnErr $? "${nodeVersionMain}"
-nodeVersionModuletest=$(getNodeVersion "${SITE}" "moduletest" "${SITES_CONFIG}")
-failOnErr $? "${nodeVersionModuletest}"
+nodeVersion=$(getNodeVersion "${SITE}" "${SITES_CONFIG}")
+failOnErr $? "${nodeVersion}"
 set -o errexit
 
 # Synchronise the sites environment repository.
-syncEnvRepo "${SITE}" "${releaseTag}" "${BRANCH}" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionMain}" "${nodeVersionMain}" "${primaryGoSubDomain}" "${secondaryGoSubDomains}" "${goRelease}"
+syncEnvRepo "${SITE}" "${releaseTag}" "${BRANCH}" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionMain}" "${nodeVersion}" "${primaryGoSubDomain}" "${secondaryGoSubDomains}" "${goRelease}"
 
 if [ "${plan}" = "webmaster" ] && [ "${BRANCH}" = "main" ]; then
-    syncEnvRepo "${SITE}" "${wmReleaseTag}" "moduletest" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionModuletest}" "${nodeVersionModuletest}"
+    syncEnvRepo "${SITE}" "${wmReleaseTag}" "moduletest" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionModuletest}" "${nodeVersion}"
 fi
--- a/infrastructure/dpladm/bin/sync-site.sh
+++ b/infrastructure/dpladm/bin/sync-site.sh
@@ -156,6 +156,25 @@ function getPhpVersion {
     exit 1
 }
 
+function getNodeVersion {
+    local branchNodeVersion nodeVersion
+
+    branchNodeVersion=$(yq eval ".sites.${1}.${2}-node-version" "${3}")
+    if [[ -n "${branchNodeVersion}" && "${branchNodeVersion}" != "null" ]]; then
+        echo "${branchNodeVersion}"
+        return
+    fi
+
+    nodeVersion=$(yq eval ".sites.${1}.node-version" "${3}")
+    if [[ -n "${nodeVersion}" && "${nodeVersion}" != "null" ]]; then
+        echo "${nodeVersion}"
+        return
+    fi
+
+    echo "Error: Neither specific or general node-version is defined for environment ${2}"
+    exit 1
+}
+
 function calculatePrimaryGoSubdomain {
     local hasGo=$(getGoRelease "${1}" "${2}")
     if [[ "${hasGo}" = "" ]]; then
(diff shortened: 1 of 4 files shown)


kasperg (approved):
:+1: LGTM

infrastructure/dpladm/bin/sync-site.sh
```diff
@@ -256,11 +275,15 @@ phpVersionMain=$(getPhpVersion "${SITE}" "main" "${SITES_CONFIG}")
...
+failOnErr $? "${nodeVersionModuletest}"
 set -o errexit
 
 # Synchronise the sites environment repository.
-syncEnvRepo "${SITE}" "${releaseTag}" "${BRANCH}" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionMain}" "${primaryGoSubDomain}" "${secondaryGoSubDomains}" "${goRelease}"
+syncEnvRepo "${SITE}" "${releaseTag}" "${BRANCH}" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionMain}" "${nodeVersionMain}" "${primaryGoSubDomain}" "${secondaryGoSubDomains}" "${goRelease}"
 
 if [ "${plan}" = "webmaster" ] && [ "${BRANCH}" = "main" ]; then
-    syncEnvRepo "${SITE}" "${wmReleaseTag}" "moduletest" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionModuletest}"
+    syncEnvRepo "${SITE}" "${wmReleaseTag}" "moduletest" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionModuletest}" "${nodeVersionMain}"
```
ITViking: shouldn't this be `nodeVersionModuletest`?

infrastructure/dpladm/bin/sync-site.sh
```diff
@@ -156,6 +156,25 @@ function getPhpVersion {
     exit 1
 }
 
+function getNodeVersion {
+    local branchNodeVersion nodeVersion
+
+    branchNodeVersion=$(yq eval ".sites.${1}.${2}-node-version" "${3}")
```
ITViking: Are we going to to use this? It is not used currently in `getPhpVersion`?

infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
```diff
@@ -15,7 +15,7 @@ ENV NEXT_TELEMETRY_DISABLED=1
 RUN yarn run build
 
 # Production image, copy all the files and run next
-FROM uselagoon/node-20:latest AS runner
+FROM uselagoon/node-${NODE_VERSION}:latest AS runner
```
ITViking: we should peg the tag and not run with `latest`. Take whatever version you have tested it with now, and peg it to that.

infrastructure/environments/dplplat01/sites.yaml
```diff
@@ -5,6 +5,7 @@ x-defaults: &default-release-image-source
   dpl-cms-release: 2026.13.0
   go-release: 2026.13.0
   php-version: 8.3
+  node-version: 20
```
ITViking, line 8: we should also test the `main` and `moduletest` version of this, if we are keeping that

ITViking (commented):
I have some comments for you to consider.

infrastructure/dpladm/bin/sync-site.sh
rasben: So right - fixed.

infrastructure/dpladm/bin/sync-site.sh
rasben: I'll be honest: I mostly just copied the getPhpVersion function, and tweaked it to match Node.


I'm not sure if this line is actually needed or not - but, if it's a functionality that is "expected" to work with PHP, it should also work with Node.

infrastructure/environments/dplplat01/sites.yaml
rasben: Added

rasben: ..I think? What do you mean by `main`? I've added `moduletest-node-version: 20`

infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
rasben: Why? We already use `:latest` in the existing code

infrastructure/dpladm/bin/sync-site.sh
rasben: Maybe @kasperg has some input?

kasperg: @ITViking  I am not sure I understand the question.

We need this feature to support different versions of node between environments e.g. Node 20 on `main`(or by default) but Node 24 on `moduletest`.

We will not have branch-specific versions of Node (or PHP) all the time but there are transition periods where they will differ. This is needed to support this.

To demonstrate that this works @rasben added both `node-version: 20` and `moduletest-node-version: 20`. Just like we did for PHP i https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/871/changes/9ce98c52ebc51ceddfba2fd538211f9fcca34023#diff-a1c41471b3b069753fb16e14837f27737f2ba59ba712b1078c55bfd28905407f.

infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
kasperg: @ITViking I see where you are coming from but I agree with @rasben here. I think pegging of the Node version is outside the scope of this PR. 

If we want to introduce pegging of images we should consider:

#### LAGOON_IMAGES_RELEASE_TAG
We have the [`LAGOON_IMAGES_RELEASE_TAG`](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/main/infrastructure/dpladm/bin/dpladm-shared.source#L5) [which is used for our PHP images](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/main/infrastructure/dpladm/env-repo-template/standard/lagoon/cli.dockerfile). It would be consistent to use this approach for Node as well.

It is worth noting that there is no `24.1.0` tag for [the `node-24` image](https://hub.docker.com/r/uselagoon/node-24/tags) - because [Node 24 was not released then](https://nodejs.org/en/about/previous-releases).

With that in mind I think the proper change here would be to:

1. Bump `LAGOON_IMAGES_RELEASE_TAG` to `26.3.0` (effectively `latest`)
2. Update `node.dockerfile` to use `LAGOON_IMAGES_RELEASE_TAG`
3. Update our release plan to include updating `LAGOON_IMAGES_RELEASE_TAG`

Alternately: Change `LAGOON_IMAGES_RELEASE_TAG` to `latest` keeping in mind that the key software we are interested in (Node, PHP etc.) is pegged to a major version through the image name.

#### Using `latest` for services
We are also tracking `latest` with [a bunch of services](https://github.com/danskernesdigitalebibliotek/dpl-platform/blob/main/infrastructure/dpladm/env-repo-template/standard/docker-compose.base-services.yml). Does this matter?

infrastructure/environments/dplplat01/sites.yaml
kasperg: As I see it this is not necessary. Branch specific versions are only necessary in situations where the two differ. I think it is preferable to only use the generic version `node-version` when that is node the case. That is also what we do for `php-version`.

The proces is:

1. A new release requires a new version of Node (or PHP)
2. Default sites get the new version of Node with the new release of Go on the production site by updating `node-version` at the same time as `go-release` in `x-defaults`.
3. Webmaster sites get the new version of Node *only on their moduletest site* by setting `moduletest-node-version`.
4. Once the webmaster sites get the newest release of Go on their production site, `node-version` is updated to the same value as `moduletest-node-version`. `moduletest-node-version` is then deleted.

infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
ITViking: @rasben to answer your question: Running with latest is a risk and you guys having a release plan ready to go presents a good opertunity to make this improvement now. 
I do however buy into @kasperg argument of it being outside of scope for this task. 
The change plan is good @kasperg :+1:

infrastructure/dpladm/bin/sync-site.sh
ITViking: We don't have any moduletest GO sites so it's not needed. Just ran the test and looked it through to see what changes were made.

infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
ITViking: @rasben to answer: There's a risk of getting a bad image in the mix when running with latest and since you guys have a change plan ready, this presented an oppertune moment to make the change. 

@kasperg I agree with your assessment and your plan for a future change.

infrastructure/environments/dplplat01/sites.yaml
ITViking: See my answer to the first code comment.

infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
kasperg: @ITViking I have prepared the first step of this plan here: 4ee23de3ff3c769dafd23b9e483f494ea1be2150. Then we are not chasing `latest`.

I suggest that we follow up with a general bump of `LAGOON_IMAGES_RELEASE_TAG` to `26.3.0` (or `latest` depending on your preference) and an update to `node.dockerfile` in a later PR. Perhaps even as a part of a separate release.

infrastructure/dpladm/bin/sync-site.sh
kasperg: With comment in https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/1084#discussion_r3074887355 in mind this change is obsolete and the conversation can be resolved.

infrastructure/environments/dplplat01/sites.yaml
kasperg: With comment in https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/1084#discussion_r3074887355 in mind this change is obsolete and the conversation can be resolved.

infrastructure/dpladm/bin/sync-site.sh
```diff
@@ -275,15 +269,13 @@ phpVersionMain=$(getPhpVersion "${SITE}" "main" "${SITES_CONFIG}")
...
 set -o errexit
 
 # Synchronise the sites environment repository.
-syncEnvRepo "${SITE}" "${releaseTag}" "${BRANCH}" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionMain}" "${nodeVersionMain}" "${primaryGoSubDomain}" "${secondaryGoSubDomains}" "${goRelease}"
+syncEnvRepo "${SITE}" "${releaseTag}" "${BRANCH}" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionMain}" "${nodeVersion}" "${primaryGoSubDomain}" "${secondaryGoSubDomains}" "${goRelease}"
 
 if [ "${plan}" = "webmaster" ] && [ "${BRANCH}" = "main" ]; then
-    syncEnvRepo "${SITE}" "${wmReleaseTag}" "moduletest" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionModuletest}" "${nodeVersionModuletest}"
+    syncEnvRepo "${SITE}" "${wmReleaseTag}" "moduletest" "${siteImageRepository}" "${siteReleaseImageName}" "${autogenerateRoutes}" "${primaryDomain}" "${secondaryDomains}" "${diskSize}" "${phpVersionModuletest}" "${nodeVersion}"
```
ITViking: I believe we wont need this one either as this is the line that will make a synchronization against the `moduletest` branch.

infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
ITViking: @kasperg nice work and good plan. I had one [correction](https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/1084#discussion_r3077515571) and then this should be releasable :+1:

infrastructure/dpladm/bin/sync-site.sh
kasperg: @ITViking OK. I have removed it.

PR #1085, opened 2026-03-26
Upgrade non-webmaster sites to run Node 24, as part of 2026.18.x

rasben:
This pull request must be merged right before we make next regular release, 2026.18.x.
We cannot merge it before hand, as that will cause problems with any hot-fix releases.

I've added a comment in sites.yaml to remember to get this merged before hand.


rasben:
https://github.com/danskernesdigitalebibliotek/dpl-web/pull/209 relies on this

PR #1099, opened 2026-04-21
Upgrade lagoon remote to 0.92.0

ITViking:
#### What does this PR do?
These are the changes to Lagoon remote chart on installing 0.92.0


ath88:
The sops keys and encrypted values are changing. Why is this happening?

ITViking:
SOPS keys are a computation based on what's in the file and is thus recalculated every time something changes. The Secrets are computed based on the keys

PR #1108, opened 2026-04-30
Deploy 2026.18.1 to editors and moduletests

xendk:
Revert the webmaster early adopters of 2026.16.0 to following the normal release cycle.


ITViking (approved):
let's go

PR #1112, opened 2026-05-05
Lagoon core upgrade to 1.48.0

ITViking:
#### What does this PR do?
This upgrades Lagoon Core to 1.48.0 and adds a runbook on how to upgrade Lagoon Core


hypesystem (approved):
Nice, good runbook, looks like something I could reasonably follow. (Haven't tested it for obvious reasons, though 😉 )

PR #1115, opened 2026-05-05
revert cron jobs back to the original setting

ITViking:
#### What does this PR do?
This should revert the scheduling of the cronjobs back to their original setting


xendk (approved):

xendk (changes requested):
Eh, too quick.

The translations jobs and the purge job only need to run once somewhere in the night.

xendk (approved):

PR #1120, opened 2026-05-07
Rework dplsh to be azure free

xendk:
Good times.


ITViking:
A few comments but that's it :smiley:

xendk:
@ITViking Think your comments got lost?

PR #1128, opened 2026-05-20
Use image from two-stage Go build

mikl:
#### What does this PR do?

Uses the new two-stage Go image build process as implemented in https://github.com/danskernesdigitalebibliotek/dpl-web/pull/599

Has already been tested for Canary deploy (thanks to @ITViking). 

https://reload.atlassian.net/browse/DDFDRIFT-514


infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
```diff
@@ -1,27 +1,33 @@
-FROM ghcr.io/danskernesdigitalebibliotek/dpl-go-node:${GO_RELEASE} as builder
+FROM ghcr.io/danskernesdigitalebibliotek/dpl-web-go:${GO_RELEASE} as builder
```
kasperg, line 1: Please consider this change.

With this approach when we deploy then all sites will switch to the new image. This requires that the defined version of said image is available. Given we merge https://github.com/danskernesdigitalebibliotek/dpl-web/pull/599 and release `2026.24.0` then `dpl-web-go:2026.24.0` will be available. However webmaster are trailing a version. They will get `dpl-web-go:2026.22.0`. This version does not exist.

For this to work we have to either:

1. Rebuild previous previous versions of Go to get `dpl-web-go:2026.22.0` etc.
2. Update our deployment handling so docker images can be templated per site
3. Reuse the image name

You might have an even better idea.

kasperg (changes requested):

mikl:
See also https://github.com/danskernesdigitalebibliotek/dpl-web/pull/599

infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
ITViking: This is very valid - we need a solution for this @mikl

ITViking (commented):
On a more general note - I would appreciate it if you could make your linter and formatter less aggressive so PR's are more honed in the actual changes

mikl:
> On a more general note - I would appreciate it if you could make your linter and formatter less aggressive so PR's are more honed in the actual changes

@ITViking good point, we solved that in one go with https://github.com/danskernesdigitalebibliotek/dpl-platform/pull/1137 :)

infrastructure/dpladm/env-repo-template/standard/lagoon/node.dockerfile
mikl: Yeah, we settled on the workaround of making a 2026.23 release for the webmasters, which will be released to them when we release 2026.24 to the other libraries.

mikl:
Two-stage build er røget med ind ifbm. 2026.23.2 releasen, så den her PR er ikke relevant mere.

PR #1140, opened 2026-06-02
upgrade lagoon core to version 1.49.0

ITViking:
#### What does this PR do?
This is an upgrade to Lagoon Core to version 1.49.0. There's nothing major to note about it


ath88 (commented):
This mostly includes changes to nats, which is disabled for us, right?

ITViking:
> This mostly includes changes to nats, which is disabled for us, right?

yes, that is correct

ath88:
approved then

PR #1143, opened 2026-06-03
add runbook on how to upgrade Lagoon Remote

ITViking:
#### What does this PR do?
It adds a runbook on how to upgrade Lagoon Remote

Read it through

### DIFF
--- a/docs/runbooks/upgrade-lagoon-remote.md
+++ b/docs/runbooks/upgrade-lagoon-remote.md
@@ -16,7 +16,7 @@ First take a backup of the API DB and the Keycloak DB and export it to a local
 machine. This will be needed if a roll-back is needed. This is due to various
 migrations Amazee runs in these databases during an upgrade.
 Run the backup and dump with `task lagoon:backup:remote-and-keycloak` which will
-export the dumps to `/infrastructure`.
+export the dumps to the platform project's `/infrastructure` folder.
 
 ## pull down the newest version of the values file
 
--- a/docs/runbooks/upgrade-lagoon-remote.md
+++ b/docs/runbooks/upgrade-lagoon-remote.md
@@ -1,0 +1,53 @@
+# When to use
+
+This runbook describes how to safely upgrade Lagoon Remote.
+
+## Prerequisites
+
+- Access to out kubernetes cluster
+- [Meld](https://meldmerge.org/)
+- [Helm](https://helm.sh/docs/intro/install/)
+- [Sops](https://getsops.io/docs/)
+- [AZ CLI](https://learn.microsoft.com/da-dk/cli/azure/install-azure-cli?view=azure-cli-latest)
+
+## Backup databases
+
+First take a backup of the API DB and the Keycloak DB and export it to a local
+machine. This will be needed if a roll-back is needed. This is due to various
+migrations Amazee runs in these databases during an upgrade.
+Run the backup and dump with `task lagoon:backup:remote-and-keycloak` which will
+export the dumps to `/infrastructure`.
+
+## pull down the newest version of the values file
+
+Go to `/lagoon-remote/`
+Run `helm show values lagoon/lagoon-remote --version x.y.z > values-x-y-z.yaml`
+
+## Decrypt the Lagoon Remote's values file
+
+The file is encrypted because it contains credentials to all of the Lagoon
+services. These can't be handled differently at this point in time.
+Decrypt the file by running: `sops -d -i values.sops.yaml`.
+This decrypts the file in place.
+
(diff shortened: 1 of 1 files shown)


docs/runbooks/upgrade-lagoon-remote.md
```diff
@@ -0,0 +1,53 @@
...
+- [Sops](https://getsops.io/docs/)
+- [AZ CLI](https://learn.microsoft.com/da-dk/cli/azure/install-azure-cli?view=azure-cli-latest)
+
+## Backup databases
+
+First take a backup of the API DB and the Keycloak DB and export it to a local
+machine. This will be needed if a roll-back is needed. This is due to various
+migrations Amazee runs in these databases during an upgrade.
+Run the backup and dump with `task lagoon:backup:remote-and-keycloak` which will
+export the dumps to `/infrastructure`.
```
ath88: Elaborate this - is `/infrastructure` a local directory?

docs/runbooks/upgrade-lagoon-remote.md
```diff
@@ -0,0 +1,53 @@
...
+services. These can't be handled differently at this point in time.
+Decrypt the file by running: `sops -d -i values.sops.yaml`.
+This decrypts the file in place.
+
+## Compare the old against the new Values file
+
+Run `meld values.sops.yaml values-x-y-z.yaml`. This opens new
+window in the meld program.
+Run through it and ensure we only accept the changes we actually want.
+Remember to pull down the changes to the `lagoon-remote.values.sops.yaml` file.
```
ath88, line 38: Consider using the 3-way-merge instead with the current version and the new version values.

docs/runbooks/upgrade-lagoon-remote.md
```diff
@@ -0,0 +1,53 @@
...
+
+Go to `/lagoon-remote/`
+Run `helm show values lagoon/lagoon-remote --version x.y.z > values-x-y-z.yaml`
+
+## Decrypt the Lagoon Remote's values file
+
+The file is encrypted because it contains credentials to all of the Lagoon
+services. These can't be handled differently at this point in time.
+Decrypt the file by running: `sops -d -i values.sops.yaml`.
+This decrypts the file in place.
```
ath88, line 31: Is it possible to do this not in-place? Is there a risk of comitting the file with unencrypted secrets, if the file is not in .gitignore?

docs/runbooks/upgrade-lagoon-remote.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Encrypt the lagoon-remote.values.sops.yaml file again
+
+Run `sops -e -i values.sops.yaml`. This encrypts the file in-place.
+
+## Helm diff the updated values file
+
+Open the `upgrade-lagoon-remote.sh` script and ensure it runs with `diff`.
+Then run the script and go over the changes. The checksum for all resources will
+have changed and there is nothing we can do about it.
```
ath88, line 48: Seems wrong to not diff by default. Can this be the other way around?

ath88 (approved):

docs/runbooks/upgrade-lagoon-remote.md
ITViking: I'm not sure I understand. What do you mean by the 3-way-merge?

docs/runbooks/upgrade-lagoon-remote.md
ITViking: It is possible to do this not i place and there is a risk of commiting the values file with unencrypted secrets. It is however not possible to upgrade the chart without encrypting the values file again

docs/runbooks/upgrade-lagoon-remote.md
ITViking: How so - we are using a script in this process, so `diff`  would have to be added or removed in any case

docs/runbooks/upgrade-lagoon-remote.md
ITViking: done

docs/runbooks/upgrade-lagoon-remote.md
ath88: So, how about a method where the un-encrypted values are written to a file in .gitignore, and when re-encrypting it is done back to the original flle? Would that be safer and guard against someone doing `git commit -a`?

docs/runbooks/upgrade-lagoon-remote.md
ath88: How about taking an input like `upgrade-lagoon-remote.sh apply`?

PR #1153, opened 2026-06-16
upgrade core and remote to a combined lagoon version 2.23.0

ITViking:
#### What does this PR do?
this is the upgrade for Lagoon to version 2.23 which includes core 1.50.0 and remote 0.96.0


ath88:
I think the alert should be defined in the cert-manager application instead, to collate the domains instead of function.

ath88 (approved):

ITViking:
> I think the alert should be defined in the cert-manager application instead, to collate the domains instead of function.

I believe you wanted to comment this on another PR ;)

